PR #2, opened 2022-10-27
Initial model serialization implementation

harubaru:
Adds a couple functions to serialize and deserialize pytorch models from binaries.

### DIFF
--- a/serializer.py
+++ b/serializer.py
@@ -2,7 +2,7 @@ import torch
 from torch import Tensor
 import numpy as np
 import tensors.tensors_pb2 as tensors_pb
-from typing import OrderedDict
+from typing import OrderedDict, Tuple, Union, BinaryIO
 
 DtypePbs = {
     torch.float32: tensors_pb.DT_FLOAT32,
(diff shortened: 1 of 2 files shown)


serializer.py
```diff
@@ -69,14 +70,78 @@ def serialize(t: Tensor) -> tensors_pb.Tensor:
...
+    f = open(filename, "wb")
+    modules = list()
+    for module_name, module in model.named_modules():
+        print(module_name)
+        attributes = list()
+        for name, param in module.named_parameters(recurse=False):
+            v = param.cpu().detach()
+            param_attr = tensors_pb.Attribute(
+                name=name,
+                tensor=tensors_pb.Tensor(
```
wbrown, line 94: Use the nice `serialize` function above? DRY up.

serializer.py
```diff
@@ -69,14 +70,78 @@ def serialize(t: Tensor) -> tensors_pb.Tensor:
...
     mv = bytearray(t.data)
     return torch.as_tensor(np.ndarray.__new__(np.memmap,
                            t.shape,
                            dtype=PbNpyDtypes[t.dtype],
                            buffer=mv,
-                           offset=0))
\ No newline at end of file
+                           offset=0))
+
+def serialize_model(model: torch.nn.Module, filename: str) -> None:
```
wbrown, line 84: This should be a writer object, not a `filename`.

serializer.py
```diff
@@ -69,14 +70,78 @@ def serialize(t: Tensor) -> tensors_pb.Tensor:
...
+                    dtype=DtypePbs[v.dtype],
+                    shape=v.shape,
+                    data=v.numpy().tobytes(),
+                    attr_type = tensors_pb.AT_PARAMETER
+                )
+            )
+            attributes.append(param_attr)
+        for name, buffer in module.named_buffers(recurse=False):
+            v = buffer.cpu().detach()
+            buffer_attr = tensors_pb.Attribute(
```
wbrown, line 104: Same comment as above about using the `serialize` function.

serializer.py
```diff
@@ -69,14 +70,78 @@ def serialize(t: Tensor) -> tensors_pb.Tensor:
...
+        )
+        modules.append(module_attr)
+    model_proto = tensors_pb.Module( # models are just modules as attributes
+        name="",
+        attributes=modules,
+    )
+    f.write(model_proto.SerializeToString())
+    f.close()
+
+def deserialize_model(model: torch.nn.Module, filename: str) -> None:
```
wbrown, line 129: This should be accepting a reader file-like object.

wbrown (changes requested):
You need to put the model example code you had in `__main__` basically as a test case. Some minor comments, and we may want to provide the user the code pattern of none_or_load, as that's an essential part of model loading.

wbrown (approved):

PR #6, opened 2023-01-19
Add Python Packaging Support

harubaru:


setup.py
```diff
@@ -0,0 +1,23 @@
...
+setup(
+    name='tensorizer',
+    version='0.0.1',
+    description='A tool for PyTorch Module, Model, and Tensor Serialization/Deserialization.',
+    url='https://github.com/coreweave/tensorizer',
+    author='CoreWeave',
+    license='MIT',
+    packages=['tensorizer'],
+    install_requires=[
+        'torch',
```
arsenetar, line 12: Should there be a version specification on torch? Probably applies to requirements file as well.

setup.py
```diff
@@ -0,0 +1,23 @@
...
+        'protobuf>=3.19.5',
```
arsenetar, line 13: I know there have been issues with newer protobuf before, perhaps we should limit this to this major version?

setup.py
harubaru: Yep, the tensorizer uses the `meta` device which is only available from 1.9.0 onwards.

setup.py
wbrown: Maybe if we regenerated stubs as part of the install? Not sure if that's what we want to do.

wbrown (approved):

PR #10, opened 2023-02-15
Fix serializer without presence of CUDA GPU

harubaru:
I got around to testing main changes in a pod and it didn't work without GPU. This PR should fix that by keeping serialization on CPU.

### DIFF
--- a/tensorizer/tensorizer.py
+++ b/tensorizer/tensorizer.py
@@ -856,13 +856,6 @@ def df_main(args: argparse.Namespace) -> None:
         args.input_directory, use_auth_token=hf_api_token
     )
 
-    cudadev = torch.cuda.current_device()
-    gb_gpu = int(
-        torch.cuda.get_device_properties(0).total_memory / (1000 * 1000 * 1000)
-    )
-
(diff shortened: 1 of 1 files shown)


tensorizer/tensorizer.py
```diff
@@ -856,13 +856,6 @@ def df_main(args: argparse.Namespace) -> None:
         args.input_directory, use_auth_token=hf_api_token
     )
 
-    cudadev = torch.cuda.current_device()
-    gb_gpu = int(
```
wbrown, line 860: You should borrow my GPU memory usage function from `finetuner.py` as it accounts for no-GPU cases rather than just tossing out this tracking.

Although, hmm, there'd be no GPU memory to measure? Might be useful for other points where we'd want it though.

wbrown (changes requested):

tensorizer/tensorizer.py
harubaru: Yeah, I think it would be useful in the validation step

wbrown (approved):

PR #17, opened 2023-03-15
Code Finalization (or close to it)

wbrown:
* Options reduced to `on_demand` and `plaid_mode`.
* By default, we use an anonymous `mmap.mmap`, and preallocate all the memory for the tensors.
* We pin the memory allocated, to reduce the amount of copies performed.


setup.py
```diff
@@ -11,18 +11,11 @@
     install_requires=[
```
Eta0, line 11: Need to add `psutil`. Unless we remove `RequestsStreamFile` (which we can now for most platforms), we may also need to add `requests`.

tensorizer/serialization.py
```diff
@@ -433,8 +457,8 @@ def read_tensors(
         except EOFError:
             return
 
-    @staticmethod
-    def _to_torch_parameter(arr: numpy.ndarray,
+    def _to_torch_parameter(self,
```
Eta0, line 460: This method can still be static.

tensorizer/serialization.py
```diff
@@ -19,15 +20,15 @@
...
-# Python doesn't know about this.
-MADV_PAGEOUT = 21
 import numpy
 import struct
+import time
 import torch
 import typing
 import logging
 import tempfile
 import regex
```
Eta0, line 30: We should switch to the built-in module `re` instead of the third-party library `regex` since our use case doesn't require the added features in `regex`.

Eta0 (changes requested):

README.md
```diff
@@ -1,6 +1,138 @@
 # tensorizer
 Module, Model, and Tensor Serialization/Deserialization
 
+## TLDR
+Extremely fast model loads from HTTP/HTTPS and S3 endpoints. GPT-J
+(`20gb`) loads at wire-speed (`~5GB/s`) on a 40gige network, and is
```
dfinster, line 6: * `20GB` is preferred instead of `20gb`
* `40GbE` is preferred instead of `40gige`

dfinster (commented):
Looks good to me with the three minor tweaks I noted.

tensorizer/serialization.py
```diff
@@ -487,29 +514,30 @@ def load_into_module(
...
-        for idx, typ, name, arr in self.read_tensors(pattern=pattern):
-            param = self._to_torch_parameter(arr, dtype, device)
+
+        for name in self.keys():
+            if pattern is not None and not pattern.match(name):
+                continue
             obj_path, attr = name.rsplit(".", 1)
             module: torch.nn.Module = modules[obj_path]
-            if typ is TensorType.PARAM:
+            entry = self._tensors[name]
```
harubaru, line 528: Could this be mistakened for `_metadata`?

harubaru: This throws an error since `_tensors` is just a normal integer.

harubaru (changes requested):

tensorizer/serialization.py
```diff
@@ -433,19 +458,21 @@ def read_tensors(
...
+                            arr: Union[numpy.ndarray, torch.nn.Parameter],
                             dtype: Optional[str] = None,
                             device=utils.get_device()) -> torch.nn.Parameter:
         """
         Convert a numpy array to a torch tensor on a device.
         """
         if dtype is not None and arr.dtype != "bool" and arr.dtype != dtype:
             arr = arr.astype(dtype)
+        if isinstance(arr, torch.nn.Parameter):
+            return arr.to(device)
```
harubaru, line 471: This turns the parameter back into a tensor (I do not know why PyTorch does this), but this removes its property of being a parameter object.

harubaru: It can be fixed by replacing this with:
```python
        if isinstance(arr, torch.nn.Parameter):
            arr.data = arr.to(device)
            if arr.grad is not None:
                arr.grad = arr.grad.to(device)
            return arr
```

PR #21, opened 2023-03-21
Fix S3 uploads and lots of bugs, reorganize files, expose custom S3 config path functionality

Eta0:
# Fixes and Reorganization
This PR contains lots of important bug fixes, small features, and reorganization for some files and method parameters.
Most commits are for a single change, so they can be browsed for more detail.

## Reorganization
- Imports from `tensorizer.serialization` can now be directly imported from `tensorizer`
  - Before: `from tensorizer.serialization import TensorDeserializer`
  - After: `from tensorizer import TensorDeserializer`
- `protobuf.py` is now included in the main `tensorizer` package
- `proto/*`, `tensors/*`, and `CMakeLists.txt` are now included in built source distributions
- Outdated `tests_*.py` files in the root directory were moved to a new `ATTIC` directory for unused code
- `tensorizer/tensors.proto` and `tensorizer/tensors_pb2.py` are now symlinks rather than copies (from `proto/` and `tensors/`, respectively)

## Features
- `stream_io.open_stream` now allows specifying an `s3_config_path` parameter to locate a particular `.s3cfg` file
  - Important when multiple config files are present or when `.s3cfg` is placed in a nonstandard location (e.g. in `/var/run`)
- `stream_io.open_stream` now allows empty-string credentials that do not trigger config parsing, for accessing public buckets that need no credentials
- `TensorDeserializer` instance methods no longer take redundant `device` and `dtype` inputs, since they are already specified in `TensorDeserializer.__init__()`
- `CURLStreamFile` now implements the context management protocol

## Tests
- There are now test cases for `open_stream`'s S3 functionality through `moto` mocked S3 calls

## Fixes
See the commits for the full list. Some highlights:
- S3 upload streams now correctly upload and clean up temporary files even when they aren't explicitly closed
  - They still *should* be closed, but they won't throw unholy errors if someone forgets
- `TensorSerializer` and `TensorDeserializer` have fewer potential errors on `__del__` under other error scenarios
- `TensorType` is now exposed as an importable name from `tensorizer.serialization` (and thus `tensorizer`)
  - It has been required for calls to `TensorSerializer.write_tensor()` but wasn't listed in `__all__`
- Lots of things were broken regarding `device` checks in `TensorDeserializer` and are now fixed (17af2fa42e6503f1741640193d0a16f8ed2feba1)
  - This includes multiple compounding bugs that prevented memory pinning from taking place when `"cuda"` was an inferred device rather than an explicitly specified string
- Method-level `filter_func` checks in `TensorDeserializer` now stack with the `__init__`-level `filter_func` check rather than replacing it
  - This had previously allowed a bug where following a more restrictive `filter_func` by a less restrictive one would lead to a buffer overflow, as the `__init__`-level `filter_func` controlled the buffer size
- `TensorDeserializer`'s `mmap` buffer is no longer eagerly closed when deserializing tensors on the CPU
  - This was resulting in use-after-free segfaults and other nastiness from deallocating memory that was still in use
- `CURLStreamFile` stream file objects are now closed correctly when `curl` exits successfully
- `open_stream` S3 endpoint URLs may now include `https://`, and are rejected if they include `http://` or another scheme
  - Previously, they were required to list no scheme, as `https://` was prepended to them unconditionally
- `open_stream` no longer allows opening local files in text mode
  - This was only allowed for local files anyway, so this usage can and should be replaced with a plain call to `open`
  - See further reasoning in the commit message: f8ec5e6d095852365316214adbec9fd4ad89d045


tests/requirements.txt
```diff
@@ -1 +1,2 @@
-transformers>=4.18
\ No newline at end of file
+transformers>=4.18
```
harubaru, line 1: Transformers broke the `PreTrainedModel` class for Tensorizer's usecase from 4.23 onwards until it was fixed in 4.27, maybe this should be set the version to 4.27.

harubaru (changes requested):
Nice fixes! Just need to update the HuggingFace serialization example to account for removing the `device` and `dtype` parameter from `load_into_module()`.

tests/requirements.txt
Eta0: Fixed! I had brought it down a while back because 4.18 is the last transformers version compatible with Python 3.6, which we were nominally supporting at the time.

PR #23, opened 2023-03-23
Fix broken Diffusers import in HuggingFace Example

harubaru:
`ModelMixin` was moved from the root directory in Diffusers to `models`, this enables the example to work with the latest Diffusers version (0.14.0).

### DIFF
--- a/examples/requirements.txt
+++ b/examples/requirements.txt
@@ -1,3 +1,3 @@
 transformers~=4.27.1
-diffusers==0.11.1
+diffusers==0.14.0
 accelerate==0.17.1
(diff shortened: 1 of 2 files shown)


examples/requirements.txt
```diff
@@ -1,3 +1,3 @@
 transformers~=4.27.1
-diffusers==0.11.1
+diffusers==0.14.0
```
NavarrePratt, line 2: Do diffusion models work for you after (de)serializing with diffusers 0.14.0?

After creating a fresh conda env with python 3.8 and pip installing the requirements, I run into a couple of issues trying to run this script.
`python hf_serialization.py stabilityai/stable-diffusion-2 ./data --model_type=diffusers --validate`
1. You also need scipy installed so it should be added to the requirements.txt:
```
ImportError: 
LMSDiscreteScheduler requires the scipy library but it was not found in your environment. You can install it with pip: `pip install
scipy`
```
2. The validation image produced after serializing/deserializing SD2 is messed up:
test

PR #26, opened 2023-05-16
feat: Add deserialization benchmarking

NavarrePratt:
This PR adds all of the files needed to run benchmarking of the (de)serialization process on CoreWeave comparing tensorizer, safetensors, and vanilla huggingface.

More info can be found in the included README.


examples/benchmark/deserialize_benchmark.py
```diff
@@ -0,0 +1,172 @@
...
+
+DEVICE = torch.device("cuda:0")
+
+
+def run_curl(output: str = "/dev/null"):
+    """
+    Used to compare pod's network speed to tensorizer's speed when using S3/HTTP
+    """
+
+    http_uri = "http://tensorized.accel-object.ord1.coreweave.com/EleutherAI/gpt-j-6B/fp16/model.tensors"
```
rtalaricw, line 44: Can we make this URL a parameter we can change?

examples/benchmark/save_models.py
```diff
@@ -0,0 +1,123 @@
...
+import torch
+from collections import defaultdict
+from pathlib import Path
+from tensorizer import TensorSerializer
+from transformers import AutoModelForCausalLM, AutoTokenizer
+from safetensors.torch import save_file
+
+MODEL_ID = os.environ.get("MODEL_ID", "EleutherAI/gpt-neo-125M")
+MODEL_PATH = Path(os.environ.get("MODEL_PATH", "./models"))
+USE_FP16 = os.environ.get("USE_FP16", False)
```
harubaru, line 14: This might be a little nitpicky but I've dealt with a similar issue before involving something like this. You should cast the environment variable to a boolean since if `USE_FP16` is the string type of `1` or `0`, the if statements using `USE_FP16` will still evaluate to `True`, except in the case where the env var is not defined it will default to `False` in this instance.

examples/benchmark/deserialize_benchmark.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+MODEL_PATH = Path(os.environ.get("MODEL_PATH", f"./models"))
+TZR_PATH = MODEL_PATH / "model.tensors"
+HF_PATH = MODEL_PATH / "hf"
+RES_PATH = Path(os.environ.get("RES_PATH", "./results"))
+NUM_TRIALS = int(os.environ.get("NUM_TRIALS", 1))
+SKIP_HF = os.environ.get("SKIP_HF", False)
+SKIP_TZR = os.environ.get("SKIP_TZR", False)
+SKIP_ST = os.environ.get("SKIP_ST", False)
```
harubaru, line 28: Same casting deal here, this can also confuse people who would want to use these env vars and the only way to actually set them to `False` is to remove the definition of these env vars from the manifest launching this script.

harubaru (changes requested):

examples/benchmark/save_models.py
Eta0: To add to that, the actually-usable way to cast a string `s` to a bool is via `s.strip().lower() not in ("", "0", "no", "f", "false")` similar to what we do in the [LLM finetuner](https://github.com/coreweave/kubernetes-cloud/blob/82a4650a688ba0bce8729d069fedce2c43177adc/finetuner-workflow/finetuner/utils.py#LL251C61-L251C61), since it handles a lot of common ways people might write that.
In this example it would look like:
```suggestion
USE_FP16 = os.environ.get("USE_FP16", "").strip().lower() not in ("", "0", "no", "f", "false")
```
Since casting via `bool(s)` has such gotchas as `bool("False")` == `True` (the only string that truly casts to `False` being the empty string).

examples/benchmark/README.md
Eta0, line 1: I have a few small edits for the `README` collected over in #28.

examples/benchmark/deserialize_benchmark.py
```diff
@@ -0,0 +1,172 @@
...
+        return
+    # Tokenize and generate
+    model.eval()
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH / "hf")
+    input_ids = tokenizer.encode(
+        "We the people, of the United States of America", return_tensors="pt"
+    ).to(DEVICE)
+
+    with torch.no_grad():
+        output = model.generate(input_ids, max_new_tokens=50, do_sample=True)
```
Eta0, line 69: The sequences generated here have nondeterministic lengths depending on *if* or *where* an EOS token is generated. Generating a shorter sequence some fraction of the time probably causes this step to run faster or slower at random. It doesn't affect the validity of the results of the benchmark because the randomness isn't biased in favour of any of the libraries—they should all be affected equally—but for a more consistent and precise measurement this should probably have a seed set so that it always produces the exact same sequence.
```suggestion
    torch.manual_seed(100)
    with torch.no_grad():
        output = model.generate(input_ids, max_new_tokens=50, do_sample=True)
```

Eta0 (changes requested):

examples/benchmark/deserialize_benchmark.py
NavarrePratt: This `run_curl()` function is left over from when I was doing benchmarks for loading from S3. It isn't being used in this code anymore so I should have removed it, but I'll remove it now.

examples/benchmark/deserialize_benchmark.py
NavarrePratt: I agree and will make the change.

However, the benchmark timing doesn't include running inference so it doesn't impact the results.

wbrown (approved):

PR #27, opened 2023-05-18
feat: Add support for bfloat16 models

NavarrePratt:
This PR adds support for models using `torch.bfloat16`.

### Overview
Tensorizer uses numpy to save the tensors as binary and numpy doesn't support `bfloat16`. This means that special conversions need to be applied which decreases the performance of both serialization and deserialization.

To be saved, the torch tensor is cast to fp32, adding an additional 16 bits. This value is converted to numpy, then bit shifted and cast down to a `np.uint16` to save the original 16 bits from torch's `bfloat16`. The reverse happens during the deserialization process.

### Benchmarks
Datapoints are averages over 5 runs and use GPT-J-6B cast to the respective precision. All were run on pods with the same CPU/RAM/GPU and were reading/writing the files from a PVC.
| Process         | fp16    | bfloat16 |
|-----------------|---------|----------|
| Serialization   | 116.63s | 172.90s  |
| Deserialization | 2.00s   | 9.25s    |

### Numpy vs Torch dtypes
Previously, the deserializer would pass around numpy arrays and dtypes. This makes things more complicated when there isn't a 1 to 1 mapping of numpy dtype to torch dtype.

For example, the "global dtype" that could be passed into the deserializer (specifying a datatype to cast everything to ignoring the metadata) was saved as a numpy datatype. Since the set of supported torch datatypes is not a 1 to 1 mapping to numpy, this was changed to be stored as a torch datatype.

The `read_tensors` function now also deals with numpy to torch tensor conversion instead of leaving that to the `_to_torch_parameter` downstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,18 @@ The `state_dict` can also be used to initialize a HuggingFace Transformers
 AutoModel. But HuggingFace Transformers performs three or more copies of
 the data, so memory use will explode.
 
+### bfloat16 Support
+
+Tensorizer supports models using the `bfloat16` data type. However, tensorizer
+uses numpy to save the tensors as binary and numpy doesn't support `bfloat16`.
+This means that special conversions need to be applied which decreases the
+performance of both serialization and deserialization.
+
+To be saved, the torch tensor is cast to fp32, adding an additional 16 bits.
+This value is converted to numpy, then bit shifted and cast down to a
+`np.uint16` to save the original 16 bits from torch's `bfloat16`. The reverse
+happens during the deserialization process.
+
 ## Running Tests
 `tensorizer` uses `unittest` for testing.
 The tests have their own set of dependencies, which can be installed with
--- a/README.md
+++ b/README.md
@@ -351,17 +351,21 @@ The `state_dict` can also be used to initialize a HuggingFace Transformers
 AutoModel. But HuggingFace Transformers performs three or more copies of
 the data, so memory use will explode.
 
-### bfloat16 Support
+### `bfloat16` Support
 
 Tensorizer supports models using the `bfloat16` data type. However, tensorizer
 uses numpy to save the tensors as binary and numpy doesn't support `bfloat16`.
-This means that special conversions need to be applied which decreases the
-performance of both serialization and deserialization.
+This means that special conversions need to be applied.
 
-To be saved, the torch tensor is cast to fp32, adding an additional 16 bits.
-This value is converted to numpy, then bit shifted and cast down to a
-`np.uint16` to save the original 16 bits from torch's `bfloat16`. The reverse
-happens during the deserialization process.
+To be saved, the torch tensor is cast to `int16` before being converted to
+numpy, which doesn't change any of the underlying data. When serialized, the
+original `blfoat16` datatype string is also saved so that it will be cast back
+to `bfloat16` during the deserialization process.
+
+The `complex32` datatype is supported in a similar way, by casting to `int32`.
+The quantized datatypes (`qint8`, `qint32`, etc.) are not currently supported
+by tensorizer as they would require supplemental quantization parameters to be
+deserialized correctly.
 
 ## Running Tests
 `tensorizer` uses `unittest` for testing.
--- a/README.md
+++ b/README.md
@@ -367,6 +367,33 @@ The quantized datatypes (`qint8`, `qint32`, etc.) are not currently supported
 by tensorizer as they would require supplemental quantization parameters to be
 deserialized correctly.
 
+### Numpy Support
+
+Tensorizer can be used with `numpy` directly to read and write
+`numpy.ndarray`'s.
+
+The serializer's `write_tensor` function handles supplying both
+`torch.Tensor`'s and `numpy.ndarray`'s.
+
+The deserializer has a separate function `read_numpy_arrays` that will return
+the data as `numpy.ndarray`'s.
+
+As explained above in [bfloat16 support](#bfloat16-support), tensorizer uses
+special conversions to write opaque datatypes, those not supported by numpy.
+Therefore, special considerations need to be taken when loading the data as
+`numpy.ndarray`'s.
+
+By default, the `read_numpy_arrays` function sets its `allow_raw_data`
+parameter to `False`. This means that if the data contains opaque datatypes,
+a `ValueError` will be raised.
+
+If you want to return the raw data regardless, set `allow_raw_data` to be
+`True`.
+
+A fifth variable is returned by the `read_numpy_arrays` generator, which is a
+`bool` that indicates whether the returned data was originally an opaque
+datatype.
+
 ## Running Tests
 `tensorizer` uses `unittest` for testing.
 The tests have their own set of dependencies, which can be installed with
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ This means that special conversions need to be applied.
 
 To be saved, the torch tensor is cast to `int16` before being converted to
 numpy, which doesn't change any of the underlying data. When serialized, the
-original `blfoat16` datatype string is also saved so that it will be cast back
+original `bfloat16` datatype string is also saved so that it will be cast back
 to `bfloat16` during the deserialization process.
 
 The `complex32` datatype is supported in a similar way, by casting to `int32`.
--- a/README.md
+++ b/README.md
@@ -367,6 +367,13 @@ The quantized datatypes (`qint8`, `qint32`, etc.) are not currently supported
 by tensorizer as they would require supplemental quantization parameters to be
 deserialized correctly.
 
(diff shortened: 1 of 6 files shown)


tensorizer/serialization.py
```diff
@@ -54,9 +55,11 @@ class TensorType(Enum):
     STATE_DICT = 2
 
 
-TENSORIZER_VERSION = 1
```
NavarrePratt, line 57: @Eta0 - Bringing the tensorizer data version convo here.

It was proposed that we still write a tensorizer data version of `1` that don't use opaque data types since those files will be backwards compatible. All files created with opaque data types will have a data version of `2` since those are not backwards compatible and needs to new code in this PR.

So to implement this we would start with writing the tensorizer version as 1 in the file. If we encounter a tensor that requires an opaque datatype, then we go back and bump the version to 2.

Next time we make a change to the data format we could get rid of this extra versioning logic and bump everything to 3 (or everything to 2). 

The counter would be that it adds too much complexity to be worth it. Bumping the data version to 2 wouldn't make any previously serialized files unsupported. It would only impact the deserialization process if you are using an older tensorizer version on a newly tensorized file.

Is this a correct summary of everything that was discussed in slack?

tensorizer/_NumpyTensor.py
```diff
@@ -0,0 +1,297 @@
...
+                something other than a ``torch.dtype``.
+            TypeError: If `self.torch_dtype` is not a string.
+        """
+        # Quick route, table lookup for common types
+        dtype = _DECODE_MAPPING.get(self.torch_dtype)
+        if dtype is not None:
+            return dtype
+        else:
+            # Long route using getattr(), any other type
+            if not self.torch_dtype:
```
NavarrePratt, line 282: @Eta0 
nit: I'm not a fan of overly indented python code, especially with smaller line lengths.  In cases like this where there's a return in the `if`, you don't technically need an `else` and can save the indentation.

Thoughts? Do you think it's more readable keeping the `else`?

Eta0: I have no preference here. Since the fact that they are two separate routes is already commented I think it would be perfectly readable unwrapping the `else`.

tensorizer/serialization.py
Eta0: Yes, that is correct. In my opinion a breaking change like that should incur bumping tensorizer to v2.0.0 if it unconditionally produces files incompatible with previous versions, and adding backwards compatibility here would be simple enough in order to not need to do that.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "tensorizer"
-version = "1.1.1"
+version = "1.2.0"
```
Eta0, line 3: This currently contains breaking changes on the code and data sides of the library, so if we're following semantic versioning, we ought to bump this to v2.0.0 with this PR.

2/3 of the significant breaking changes could be made compatible by:
- Keeping the `_convert_dtype_to_torch` function to allow `numpy` dtype arguments for the deserializer, where valid, and
- Writing `TENSORIZER_VERSION` as 1 in compatible serialized files (as discussed in other comments).

The only one that seems mostly unavoidable is altering the return value type of `TensorDeserializer.read_tensors()` from `numpy.ndarray`s to `torch.Tensor`s (at least when opaque dtypes are present). `read_tensors()` returning tensors is more sensible anyway, but it does change the interface a small bit.

tensorizer/serialization.py
```diff
@@ -190,7 +159,7 @@ def __init__(
         ],
         device: Union[torch.device, str, None] = None,
         filter_func: Optional[Callable[[str], Union[bool, Any]]] = None,
-        dtype: Union[numpy.dtype, str, None] = None,
+        dtype: torch.dtype = None,
```
Eta0, line 162: It's important to annotate that this parameter is optional.
```suggestion
        dtype: Optional[torch.dtype] = None,
```

tensorizer/serialization.py
```diff
@@ -609,7 +574,18 @@ def read_tensors(
...
+                numpy_dtype, *torch_dtype = dtype.split(OPAQUE_DTYPE_SEP)
+                if len(torch_dtype) == 0:
+                    torch_dtype = None
+                elif len(torch_dtype) == 1:
+                    torch_dtype = torch_dtype[0]
+                else:
+                    raise ValueError("Can't deserialize a tensor with "
+                                     "multiple opaque dtype separators "
+                                     f"({OPAQUE_DTYPE_SEP}) in its dtype: "
+                                     f"{dtype}")
```
Eta0, line 588: Including a literal null byte in a string meant to be read by a human might lead to weird things. We should use its `repr` instead so it formats as `'\x00'`:
```suggestion
                                     f"({OPAQUE_DTYPE_SEP!r}) in its dtype: "
                                     f"{dtype!r}")
```

Eta0 (changes requested):

pyproject.toml
NavarrePratt: From a discussion with @wbrown on slack, there is still user desire to deal with numpy as input/output types. The serializer already does this by `write_tensors` allowing both numpy ndarrays and torch tensors as input. To keep the same "single function, multiple in/outputs" pattern, I'll add a `as_numpy`  flag to `read_tensors()` that controls what dtype is returned.

This doesn't change the fact that this PR changes the default behavior and is therefore a breaking change, but it still gives a path for the user to interact with tensorizer only through numpy ndarray's.

They won't be able to use the "global dtype" in the deserializer's constructor to cast the numpy arrays but that was always only done when registering it to the torch module.

tests/test_serialization.py
```diff
@@ -202,15 +233,20 @@ def test_plaid_mode_guards(self):
 
         deserialized.close()
 
-    def test_s3(self):
+    @patch("tensorizer.stream_io._infer_credentials")
+    def test_s3(self, credentials_mock):
+        credentials_mock.return_value = s3_credentials
```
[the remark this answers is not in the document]
NavarrePratt: I like this, much cleaner. I made the change :)

tensorizer/_NumpyTensor.py
```diff
@@ -277,21 +277,21 @@ def _decode_torch_dtype(self) -> torch.dtype:
...
+        module, *dtype_name = self.torch_dtype.split(".", 1)
+
+        # Ensure that it's actually "torch.something"
+        if module != "torch" or len(dtype_name) != 1:
+            raise ValueError(f"Invalid torch_dtype: {self.torch_dtype}")
+
+        dtype = getattr(torch, dtype_name[0])
+        # Ensure that it's a real dtype
+        if dtype is None or not isinstance(dtype, torch.dtype):
+            raise ValueError(f"Invalid torch_dtype: {self.torch_dtype}")
```
Eta0, line 295: Slight bug in my code here. I think this was intended to raise a `ValueError` if `torch.<whatever>` doesn't exist, but instead it throws an `AttributeError` on `getattr` since no default is defined. The `AttributeError` is actually pretty informative, so the below change adds it to the traceback of the intended `ValueError` and makes it distinct from the wrong-type case, which is given a more descriptive error message as well.
```suggestion
        try:
            dtype = getattr(torch, dtype_name[0])
            # Ensure that it's a real dtype
            if not isinstance(dtype, torch.dtype):
                raise TypeError(
                    "Provided torch_dtype is not an instance of torch.dtype"
                    f" (type: {type(dtype).__name__})"
                )
        except (AttributeError, TypeError) as e:
            raise ValueError(f"Invalid torch_dtype: {self.torch_dtype}") from e
```

Eta0 (changes requested):
A couple tiny changes with prewritten fixes (these could be put into a PR if desired).

harubaru (changes requested):
Tested the changes with various GPT models using the HuggingFace example with the `bfloat16` dtype and the code seems to work flawlessly. Only issue I ran into is HuggingFace Transformers not being able to load models without an author but that's a HuggingFace issue.

README.md
```diff
@@ -351,6 +351,22 @@ The `state_dict` can also be used to initialize a HuggingFace Transformers
...
 
+### `bfloat16` Support
+
+Tensorizer supports models using the `bfloat16` data type. However, tensorizer
+uses numpy to save the tensors as binary and numpy doesn't support `bfloat16`.
+This means that special conversions need to be applied.
+
+To be saved, the torch tensor is cast to `int16` before being converted to
+numpy, which doesn't change any of the underlying data. When serialized, the
+original `blfoat16` datatype string is also saved so that it will be cast back
```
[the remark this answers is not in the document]
NavarrePratt: Nice catch, pushed the fix.

wbrown:
@Eta0 @NavarrePratt Are we done with this yet?

Eta0 (approved):
Good to merge!

PR #34, opened 2023-08-02
(feat) Tensorizer Hash Verification and Live Model Verification

wbrown:
Implements hash verification of tensors, and also the ability to verify a model that's been loaded in memory.

`tensorizer` had already had hashing of tensors implemented in the `TensorSerializer`, and we actually already read the hash data in as part of `TensorDeserializer.` The missing piece was to reconstruct the data structure (without the hashes) so that the state matches exactly.

On load of tensors, we store the raw representation of the headers on the `TensorDeserializer` object. Providing `verify_hashes=True` will cause the tensors to be hashed and compared to the hash values recorded. Failures will raise a `ValueError` exception.

A bonus feature that actually arose from a customer need is being able to verify that the model data loaded matches what we expect from `tensorizer`. To rule out sunspot activity on non-ECC GPUs, you know?

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -85,7 +85,7 @@ def check_deserialized(
     allow_subset: bool = False,
 ):
     orig_sd = model_digest(model_name)
-
+    
     if not allow_subset:
         test_case.assertEqual(
             orig_sd.keys(),


harubaru (changes requested):

tensorizer/serialization.py
```diff
@@ -526,10 +533,69 @@ def _decode_hashes(b: bytes) -> List[TensorHash]:
...
+                        f"Expected {hash_crc}, got {crc}."
+                    )
+            elif hash['type'] == HashType.SHA256:
+                sha = hashlib.sha256(headers)
+                sha.update(mv)
+                sha_digest = sha.digest()
+                if sha_digest != hash['hash']:
+                    raise RuntimeError(
+                        f"Tensor '{name}' failed SHA256 verification. "
+                        f"Expected {hash['hash']}, got {sha}."
```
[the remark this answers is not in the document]
wbrown: Good catch. Fixed.

harubaru (approved):

tensorizer/stream_io.py
```diff
@@ -137,6 +138,7 @@ def __init__(
             "Accept-Encoding: identity",
             "-s",
             "-f",
+            "--raw",
```
Eta0, line 141: This flag will lead to incorrect behaviour with a chunked `Transfer-Encoding`.
`Accept-Encoding: identity` only affects `Content-Encoding`, but `--raw` affects interpretation of both.
```suggestion
```

tensorizer/serialization.py
```diff
@@ -616,25 +684,23 @@ def _read_numpytensors(
...
                 # Check if the name is in our pre-filtered list of keys
                 # from the class-level filter_func, and then verify
                 # that it passes the method-level filter_func.
                 # Skip it if it fails either check.
                 if name not in self.keys() or (
-                    filter_func is not None and not filter_func(name)
+                        filter_func is not None and not filter_func(name)
                 ):
                     self._file.seek(data_length, io.SEEK_CUR)
                     continue
```
Eta0, line 706: Reading hashes needs to be done after the filter check, as `self._metadata[name]` won't exist for pre-filtered entries.

Eta0 (changes requested):

tensorizer/serialization.py
```diff
@@ -719,10 +795,11 @@ def read_tensors(
...
-        self,
-        filter_func: Optional[Callable[[str], Union[bool, Any]]] = None,
-        num_tensors: int = -1,
-        allow_raw_data: bool = False,
+            self,
+            filter_func: Optional[Callable[[str], Union[bool, Any]]] = None,
+            num_tensors: int = -1,
+            allow_raw_data: bool = False,
+            verify_hash: Union[bool, None] = None,
     ) -> Iterator[Tuple[int, int, str, numpy.ndarray, bool, Optional[str]]]:
```
Eta0, line 803: This function needs to have its docstring updated to explain its new `verify_hash` parameter.

tensorizer/serialization.py
```diff
@@ -903,7 +988,7 @@ def load_into_module(
             elif entry["type"] is TensorType.STATE_DICT:
                 raise NotImplementedError(
                     "This was serialized using the write_state_dict() method,"
-                    " and cannot be loaded using the load_tensors() method."
+                    " and cannot be loaded using the load_into_tensors() method."
```
Eta0, line 991:
```suggestion
                    " and cannot be loaded using the load_into_module() method."
```

tests/test_serialization.py
```diff
@@ -319,3 +323,41 @@ def custom_check(tensor_name: str) -> bool:
...
+                del orig_sd
+                try:
+                    with open(serialized_model, "rb") as in_file:
+                        deserialized = TensorDeserializer(in_file, device="cpu")
+                        deserialized.close()
+                        model_to_verify = AutoModelForCausalLM.from_pretrained(
+                            model_name).to(device)
+                        deserialized.verify_module(model_to_verify)
+                        del model_to_verify
+                        del deserialized
```
Eta0, line 361: This currently deserializes to the CPU instead of the intended device, and closes the deserializer too early.
```suggestion
                    with open(serialized_model, "rb") as in_file:
                        deserialized = TensorDeserializer(
                            in_file, device=device
                        )
                        model_to_verify = AutoModelForCausalLM.from_pretrained(
                            model_name
                        ).to(device)
                        deserialized.verify_module(model_to_verify)
                        deserialized.close()
                        del model_to_verify, deserialized
```

tensorizer/serialization.py
```diff
@@ -912,6 +997,32 @@ def load_into_module(
...
+            # Check if the module has the attribute
+            if not hasattr(module, attr):
+                raise ValueError(
+                    f"Module {obj_path} does not have attribute {attr}"
+                )
+            numpy_tensor = _NumpyTensor.from_tensor(getattr(module, attr))
+            self._verify_hashes(name,
+                                entry["hashes"],
+                                entry["raw_headers"],
+                                numpy_tensor.data.tobytes())
```
Eta0, line 1025: Two concerns:
1. This does a one-way comparison, so it passes as long as the `TensorDeserializer` is a subset of `m`.
    1. This means even an empty `TensorDeserializer` would pass this check
    2. This needs a supplementary check that `self.keys() == m_keys` where `m_keys = m.state_dict().keys() | dict(m.named_buffers()).keys()`
    3. Like for `check_deserialized` in `tests/test_serialization.py`, this gets ambiguous when the `TensorDeserializer` is filtered by a `filter_func`, which can be solved in two ways:
        1. Add an `allow_subset` flag that uses the current behaviour, *and/or*
        2. Add a `filter_func` parameter to `verify_module` that lets someone choose to only compare against a specific subset of `m_keys`
2. A function that has no observable effect or return value but sometimes raises an exception seems less useful than one that returns a `bool`, since with a bool you can choose to e.g. `assert verify_module(...)` or use it in an if-statement.

Eta0 (changes requested):

tensorizer/serialization.py
wbrown: This has been addresssed.

tensorizer/serialization.py
wbrown: 1. I don't think we can easily compare tensors for the absence of a tensor in the module, as listing the named modules yields things that don't quite 100% match. I'll need to noodle on this. We'll add a `TODO` for this.
2. The function has been revamped to yield a boolean and a list of tuples of name, bool indicating the status of the individual tensors. A test case has been added for a tensor that has changed.

tensorizer/serialization.py
```diff
@@ -581,12 +586,12 @@ def _verify_hashes(name: str,
...
                 sha_digest = sha.digest()
                 if sha_digest != hash['hash']:
-                    raise RuntimeError(
+                    raise HashMismatchError(
                         f"Tensor '{name}' failed SHA256 verification. "
                         f"Expected {hash['hash']}, got {sha_digest}."
                     )
             else:
-                raise RuntimeError(
+                raise HashMismatchError(
```
Eta0, line 594: Not sure if this quite qualifies as a `HashMismatchError` like the other two. It's probably more accurately described with a `ValueError` like other parsing & `.tensors` version errors.

Eta0 (changes requested):
Left one comment on a choice of exception type. My PR #37 fixes all the other issues I noticed.

tensorizer/serialization.py
Eta0: [`check_deserialized` in the test suite](https://github.com/coreweave/tensorizer/blob/0478eed41fa14a874ac82bcaa06adf4def2eca24/tests/test_serialization.py#L41-L48) implements such a comparison between the keys in a deserialized module versus a reference module. It should be usable as a base for `verify_module` as well:

```python
def check_deserialized(deserialized, model_name: str, allow_subset=False):
    orig_model = AutoModelForCausalLM.from_pretrained(model_name)
    orig_sd = orig_model.state_dict()
    # Non-persistent buffers are serialized in tensorizer,
    # but aren't included in a state_dict() in PyTorch.
    orig_sd.update(orig_model.named_buffers())
    if not allow_subset:
        assert orig_sd.keys() == deserialized.keys()
    ...
```

`state_dict() | named_buffers()` (and potentially some other unions) seem to give a comprehensive set of entries, but it'd need to be tested on more models than `gpt-neo-125M` to make sure of that.

wbrown: Hm. What's the overhead of doing that, though?

tensorizer/serialization.py
wbrown: Agreed, changed to `ValueError`.

PR #37, opened 2023-08-09
fix: Hash Verification Fixes & Polish

Eta0:
# Hash Verification Fixes & Polish
This PR contains fixes and updates for #34.

## Fixes
- Fixed `load_into_module` ignoring its `verify_hash` parameter
- Fixed a crash in `load_into_module` when a deserialized submodule is not present in the reference module
- Fixed a leaked `memoryview` on failed hash verification causing errors in `TensorDeserializer.close()` and `TensorDeserializer.__del__()`

## Tests:
- Added `test_verification_fail` to verify `verify_hash` failures outside of `TensorDeserializer.verify_module()`
- Added additional checks to `test_module_verification_fail` to ensure it caught the correct modified submodule
- Switched from `accel-object` to the `object` endpoint when reading from the `s3://tensorized` bucket during test cases
- Switched from `assert` statements to more descriptive `unittest.TestCase.assert*` methods
- Added clearer error messages for various test case failures
- Optimized speed of test cases
  - Deduplicated work serializing modules and loading reference models in `TestVerification`
  - Modified `check_deserialized()` to compare against cached hashes of reference models instead of reloading them on every call

## Docs
- Added missing docstring parameters and elaborated on existing ones
- Added `Args:` and `Returns:` blocks where missing
- Added `Raises:` blocks
- Simplified some type hints, corrected typos

## Misc.
- Restored `black` formatting
- Updated `black` version in pre-commit hooks
- Elided unnecessary tensor data copy in `verify_module()`
- Simplified and optimized handling of `bytes` objects in header parsing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^.*_pb2\.py|ATTIC/.*$
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.3.0
+    rev: 23.7.0
     hooks:
       - id: black
         language_version: python3.10
--- a/examples/benchmark/deserialize_benchmark.py
+++ b/examples/benchmark/deserialize_benchmark.py
@@ -1,11 +1,13 @@
-import torch
-import shutil
 import os
+import shutil
 import time
 from pathlib import Path
+
+import numpy as np
+import torch
+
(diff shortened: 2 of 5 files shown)


wbrown (approved):
Good changes. :+1:

PR #39, opened 2023-08-10
fix: Add missing `verify_hash` parameter to `read_tensors()`

Eta0:
# Hash Verification Control in `TensorDeserializer.read_tensors()`

This change adds a `verify_hash` parameter to `TensorDeserializer.read_tensors()` and bumps the code version to v2.1.1.

The nearly-identical function `TensorDeserializer.read_numpy_arrays()` had this parameter added in #34, and `read_tensors()` was intended to mirror that addition, but it was left out.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -1109,8 +1109,6 @@ class TensorDeserializer(collections.abc.Mapping):
                 this function prior to loading the tensors into a module.
                 If ``lazy_load=False``, this error case is impossible.
         """
-        # TODO: Account for the case where the module has more tensors than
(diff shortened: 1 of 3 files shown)


tensorizer/serialization.py
```diff
@@ -1102,8 +1111,9 @@ def verify_module(
         """
         # TODO: Account for the case where the module has more tensors than
```
wbrown, line 1112: We should delete this TODO note, as that's fixed now.

wbrown (changes requested):

wbrown (approved):

PR #44, opened 2023-09-06
(feat)Add request header to `CURLStreamFile` along with additional metrics.

wbrown:
This adds response headers to the `CURLStreamFile `class by using curl's `--dump-header /dev/stderr` option, and reading from `stderr=subprocess.PIPE`. The bonus is that we can detect when we did not get a `HTTP 1.1 2*` response and handle it accordingly.

Also adds a few more tracking attributes:
* `popen_latencies`: A list of the time it took to start the cURL process.
* `http_response_latencies`: A list of the time it took to get the first HTTP response from the server.
* `response_headers`: A dictionary of the HTTP response headers.
* `bytes_read`: The number of bytes read from the stream.
* `bytes_skipped`: The number of bytes skipped from the stream.
* `read_operations`: The number of read operations performed on the stream.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -442,11 +442,11 @@ class TensorDeserializer(collections.abc.Mapping):
     # If our _file object has 'response_headers' attribute, we can use it
     # to determine if we were cached or not.
     @property
-    def cache_status(self) -> Union[bool, str]:
+    def cache_status(self) -> Optional[str]:
         if hasattr(self._file, "response_headers"):
-            return self._file.response_headers.get("x-cache-status", False)
+            return self._file.response_headers.get("x-cache-status", None)
         else:
-            return False
+            return None
 
     def __getitem__(self, name) -> torch.nn.Parameter:
         if self._plaid_mode:
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -439,13 +439,12 @@ class TensorDeserializer(collections.abc.Mapping):
         else:
             return self._file.tell()
 
-    # If our _file object has 'response_headers' attribute, we can use it to
-    # to get the total size of the file, in addition to whether the object
-    # was cached.
+    # If our _file object has 'response_headers' attribute, we can use it
+    # to determine if we were cached or not.
     @property
-    def is_cached(self) -> bool:
+    def cache_status(self) -> Union[bool, str]:
         if hasattr(self._file, "response_headers"):
-            return self._file.response_headers.get("X-Cache-Status", "MISS") == "HIT"
+            return self._file.response_headers.get("X-Cache-Status", False)
         else:
             return False
 
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -444,7 +444,7 @@ class TensorDeserializer(collections.abc.Mapping):
     @property
     def cache_status(self) -> Union[bool, str]:
         if hasattr(self._file, "response_headers"):
-            return self._file.response_headers.get("X-Cache-Status", False)
+            return self._file.response_headers.get("x-cache-status", False)
         else:
             return False
 
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -432,10 +432,22 @@ class TensorDeserializer(collections.abc.Mapping):
 
     @property
     def total_bytes_read(self) -> int:
+        if hasattr(self._file, "bytes_read"):
+            return self._file.bytes_read
         if self._file.closed:
             return self.total_tensor_bytes
         else:
             return self._file.tell()
+
+    # If our _file object has 'response_headers' attribute, we can use it to
+    # to get the total size of the file, in addition to whether the object
+    # was cached.
+    @property
+    def is_cached(self) -> bool:
+        if hasattr(self._file, "response_headers"):
+            return self._file.response_headers.get("X-Cache-Status", "MISS") == "HIT"
+        else:
+            return False
 
     def __getitem__(self, name) -> torch.nn.Parameter:
         if self._plaid_mode:
--- a/tensorizer/stream_io.py
+++ b/tensorizer/stream_io.py
@@ -202,7 +202,7 @@ class CURLStreamFile:
         self.popen_latencies.append(popen_end - popen_start)
         self.http_response_latencies.append(resp_begin - popen_end)
 
-        if not resp.startswith(b"HTTP/1.1 2") or not resp.startswith(b"HTTP/2 2"):
+        if not resp.contains(b"HTTP/1.1 2") and not resp.startswith(b"HTTP/2 2"):
             self._curl.terminate()
             self._curl.wait()
             raise IOError(f"Failed to open stream: {resp.decode('utf-8')}")
--- a/tensorizer/stream_io.py
+++ b/tensorizer/stream_io.py
@@ -6,10 +6,11 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import time
 import typing
 import weakref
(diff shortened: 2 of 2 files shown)


tensorizer/stream_io.py
```diff
@@ -162,13 +174,52 @@ def __init__(
...
+        if not hasattr(self, "bytes_read"):
+            self.bytes_read: int = 0
+        if not hasattr(self, "bytes_skipped"):
+            self.bytes_skipped: int = 0
+        if not hasattr(self, "read_operations"):
+            self.read_operations: int = 0
+        self.popen_latencies.append(popen_end - popen_start)
+        self.http_response_latencies.append(resp_begin - popen_end)
+
+        if not resp.startswith(b"HTTP/1.1 2"):
```
robbat2: safety: this will break when we start returning HTTP/2 responses. leave a comment and fix later

tensorizer/stream_io.py
```diff
@@ -146,6 +156,8 @@ def __init__(
             curl_path,
             "--header",
             "Accept-Encoding: identity",
+            "--dump-header",
+            "/dev/stderr",
```
robbat2, line 160: nit: re-using stderr for headers is slightly problematic because you might not get the errors clearly, or they might be mixed with the headers. Fix later, use a seperate FD or tempfile for the header

robbat2 (approved):
good enough, few things to improve later

tensorizer/stream_io.py
wbrown: Already handled, as it blows up on the test cases.

tensorizer/serialization.py
```diff
@@ -432,11 +432,23 @@ def _read_shapes(obj, num_elems) -> List[int]:
...
         else:
             return self._file.tell()
 
+    # If our _file object has 'response_headers' attribute, we can use it to
+    # to get the total size of the file, in addition to whether the object
+    # was cached.
+    @property
+    def is_cached(self) -> bool:
+        if hasattr(self._file, "response_headers"):
+            return self._file.response_headers.get("X-Cache-Status", "MISS") == "HIT"
```
robbat2, line 448: there are OTHER values beyond HIT/MISS, can you please capture & expose them?

tensorizer/stream_io.py
```diff
@@ -162,13 +174,52 @@ def __init__(
...
+        if not hasattr(self, "popen_latencies"):
+            self.popen_latencies: List[float] = []
+        if not hasattr(self, "http_response_latencies"):
+            self.http_response_latencies: List[float] = []
+        if not hasattr(self, "bytes_read"):
+            self.bytes_read: int = 0
+        if not hasattr(self, "bytes_skipped"):
+            self.bytes_skipped: int = 0
+        if not hasattr(self, "read_operations"):
+            self.read_operations: int = 0
```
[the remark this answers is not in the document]
wbrown: Addressed in this PR anyway, easy enough to do.

tensorizer/stream_io.py
```diff
@@ -162,13 +174,52 @@ def __init__(
...
+            self._curl.terminate()
+            self._curl.wait()
+            raise IOError(f"Failed to open stream: {resp.decode('utf-8')}")
+        # Make stderr non-blocking so that we can read the rest of the header without
+        # blocking the main thread.
+        os.set_blocking(self._curl.stderr.fileno(), False)
+        # Read the rest of the header response and parse it.
+        header_lines = self._curl.stderr.readlines()
+        self.response_headers = {}
+        for line in header_lines:
```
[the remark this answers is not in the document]
wbrown: I've added a check for `:` in the loop, which should mitigate this issue slightly?

tensorizer/stream_io.py
wbrown: Added a `TODO`.

tensorizer/serialization.py
wbrown: Hm. They're indirectly available with `response_headers`. Do we want to return a string instead of a `boolean`?

robbat2: yes please. I suspect it will be more important on the problematic requests than the requests that are working well. (e.g. why did it miss, was it revalidating etc)

tensorizer/stream_io.py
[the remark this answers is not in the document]
robbat2: yes, good enough for now, we'll fix it later when stderr and the headers go to different pipes/dests.

tensorizer/stream_io.py
```diff
@@ -162,17 +174,62 @@ def __init__(
...
+            line = line.decode("utf-8").strip()
+            if not line or ':' not in line:
+                continue
+            k, v = line.split(":", maxsplit=1)
+            self.response_headers[k.strip()] = v.strip()
```
robbat2, line 213: two subtle bug I just caught here:
1. multiple headers with the same name are *valid* in HTTP, and have special concatenation rules. [most headers should be concatenated with commas, but there are exceptions].
2. the header names are not case sensitive.

maybe a good enough solution is to just alwatys concat here, with commas.

tensorizer/serialization.py
wbrown: OK, changing to `cache_status` that returns either `False` or the string value. That good?

tensorizer/stream_io.py
wbrown: Good catch. I'd say that starts falling into the realm of 'just use a HTTP parsing lib, dummy'. I'll add a` FIXME` note.

wbrown: `FIXME` added, and lowercased header keys.

Eta0: Fixed this by using Python's built-in `http.client.parse_headers` routine in 558176610d5b159b7bb7cb97304992f99e511b4f

tensorizer/stream_io.py
Eta0: Fixed this by using a separate pipe in 558176610d5b159b7bb7cb97304992f99e511b4f and writing to it through `/dev/fd`

tensorizer/serialization.py
```diff
@@ -432,11 +432,22 @@ def _read_shapes(obj, num_elems) -> List[int]:
...
             return self._file.tell()
 
+    # If our _file object has 'response_headers' attribute, we can use it
+    # to determine if we were cached or not.
+    @property
+    def cache_status(self) -> Union[bool, str]:
+        if hasattr(self._file, "response_headers"):
+            return self._file.response_headers.get("x-cache-status", False)
+        else:
+            return False
```
Eta0, line 449: This would be better as an `Optional[str]`
```suggestion
    @property
    def cache_status(self) -> Optional[str]:
        if hasattr(self._file, "response_headers"):
            return self._file.response_headers.get("x-cache-status", None)
        else:
            return None
```

tensorizer/stream_io.py
```diff
@@ -162,17 +176,54 @@ def __init__(
...
+            self._curl.wait()
+            raise IOError(f"Failed to open stream: {resp.decode('utf-8')}")
+        # Make the header pipe non-blocking so that we can read the rest
+        # of the header without blocking the main thread.
+        os.set_blocking(header_r_pipe, False)
+        # Read the rest of the header response and parse it.
+        # noinspection PyTypeChecker
+        self.response_headers = http.client.parse_headers(self._header_pipe)
+
+        self.error_buffer = ""
```
Eta0, line 213: This is unused
```suggestion
```

Eta0 (changes requested):
Minor changes, otherwise good

robbat2 (approved):

PR #46, opened 2023-09-09
perf(serialization): Optimize bulk serialization through concurrency

Eta0:
# Bulk Serialization Optimizations

This change increases performance on `TensorSerializer.write_module()` by about 3x, and even more on CUDA modules.
Serializing a CPU module is always (and always has been) faster than serializing a CUDA module, but this change hides several latencies and narrows this speed gap.

## Latency Hiding

When writing many tensors as an atomic operation—e.g. in a call to `write_module`—it is not necessary for an individual entry to be complete (both tensor and header written) before work begins on the next tensor **as long as all tensors and all headers are written before the `write_module` call returns.** This fact opens up more possibilities for optimization when analyzing latencies between **separate tensor writes.**

In fact, several classes of latency slow down the previous implementation more than throughput limitations.

For reference, the previous implementation's ordering of operations was:

- Transfer a GPU tensor to the CPU (for CUDA tensors)
- Hash a single tensor, and write it to a file at the same time
- Write the header for that tensor
- Proceed to the next tensor

Here are the inefficiencies in that approach:

- GPU-to-CPU `memcpy`
  - As CUDA tensors are transferred to the CPU as-needed, immediately before each tensor write, all serialization steps are stalled (with no work being done) while waiting for each call to `tensor.cpu()` from start to end
  - Mainly GPU resources are occupied by this task
  - No other steps of serialization involve the GPU, so almost any other task *could* be worked on while waiting for this
  - A tensor must nonetheless be transferred to the CPU as a prerequisite for all other tasks on *that particular tensor*, so optimizations can only apply here during bulk processing
- Unnecessary synchronization between **hash computations** and **writing tensors to a file**
  - Hash computations are entirely CPU-bound
    - Their only prerequisite is having access to tensor data on the CPU (addressed above)
    - As soon as a tensor arrives on the CPU, it is ready to be hashed
  - Writing tensors to a file is primarily IO-bound
  - The implementation before this change only computes the pair of hashes for a single tensor, and then waits for that tensor to be written to the file—even though those tasks have no dependence on each other
  - All CPU cores could be used to compute hashes for different tensors concurrently if not for this coupling
  - Latency is thus introduced by underutilization of CPU cores
- Unnecessary synchronization between **writing tensors to a file** and **writing headers to a file**
  - The previous implementation required both of these operations to complete before proceeding to any work on the next tensor
  - Writing the header for a tensor to a file only has a dependency on the hash computation finishing for that tensor, which is unrelated to the tensor itself being written to a file
    - This can lead to latency being introduced from I/O being underutilized, since subsequent tensor writes may be forced to wait indirectly on previous hash computations by waiting on this task

## Optimizations

Analyzing the latencies identified above leads to a parallelized 4-pipeline approach that minimizes unnecessary task dependencies and fully utilizes each individual resource.

- Pipeline 1: Device-to-host transfers
  - Dependencies: *none*
  - Concurrency: 1 dedicated thread
- Pipeline 2: Computation (hashes)
  - Dependencies: pipeline 1
  - Concurrency: thread pool, sized according to hardware concurrency
- Pipeline 3: Basic file I/O
  - Dependencies: pipeline 1
  - Concurrency: thread pool, 1 to 4 workers
    - Makes use of `os.pwrite` when available, allowing for some degree of true parallelism and ultimately speeding up I/O
    - Uses 1 worker when `os.pwrite` is not available, since writes must then synchronize on control of a single file offset
- Pipeline 4: Header file I/O
  - Dependencies: pipeline 1, pipeline 2
  - Concurrency: thread pool, 1 to 4 workers
    - Same principles as the basic file I/O thread pool
    - Shares a lock on the file offset with the basic I/O thread pool when relevant
    - **Does not** share a work queue or threads with the basic file I/O thread pool

These are all in addition to the main process thread, which does header formatting, queues tasks for all of the other pipelines, and performs some metadata tracking.

This change also reorganizes metadata tracking to stay in memory whenever possible, and only updates file contents when actually necessary.

## Future work

The header file I/O thread pool is less efficient than it could be because it may wait without doing any work if computation tasks finish far enough out-of-order that it **A)** doesn't notice them and **B)** continues to wait on earlier unfinished computation tasks, doing nothing. This is isn't a big issue because file I/O is usually still being utilized by the basic file I/O thread pool even when the header file I/O thread pool isn't actively writing anything.

<details>
<summary><h2>Good vibes!</h2></summary>

⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣼⣷⣄⡀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢰⣾⣚⣿⠾⠿⠶⠤⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣠⠤
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⣀⣀⣀⣤⣤⣤⣤⣤⣤⣤⣤⣤⣄⣀⡀⠀⠀⠠⠤⣾⢷⣿⣿
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⣤⣴⣶⣾⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣷⣶⣦⣄⣀⠉⠑⠻
⠀⠀⠀⠀⠀⠀⠀⢀⣠⣴⣿⣿⣿⣿⣿⣿⣿⠟⣻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣶⣄
⠀⠀⠀⠀⠀⢀⣴⣿⣿⣿⣿⣿⣿⣿⣿⡿⢁⣼⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⠀⠀⠀⠀⣰⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠁⣼⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⠀⠀⠀⣨⣽⣿⣿⣿⣿⣿⣿⣿⣿⠟⢁⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠋⣀⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿
⠀⠀⣸⣿⣿⣿⣿⣿⣿⣿⣿⠟⢁⣴⣿⡆⠹⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠟⢁⣴⣿⡄⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿
⠀⠀⣿⣿⣿⣿⣿⣿⣿⡿⠁⣠⣿⠟⢿⣿⣄⠹⣿⣿⣿⣿⣿⣿⣿⠟⢁⣠⣿⡏⠀⢽⣆⠻⣿⣿⣿⣿⣿⣿⣿⠃
⠀⠀⣿⣿⣿⣿⣿⣿⠋⢠⣾⣿⡏⠀⢸⣿⣿⡄⠹⣿⣿⣿⡿⠟⢁⣴⣿⣿⣿⠀⠀⣾⣿⣆⠙⣿⣿⣿⣿⣿⡟⢀
⠀⠀⢸⣿⣿⣿⠟⠁⢀⣾⣿⣿⠀⠀⢸⣿⣿⣿⣄⠘⣿⠋⣠⣴⣿⣿⣿⣿⡟⠀⠀⣿⣿⣿⣧⠘⢿⣿⣿⣿⠀⣼
⠀⠀⠀⢿⣿⠏⠀⠀⣼⣿⣿⣿⠀⠀⢿⣿⣿⣿⣿⣦⡀⣴⣿⣿⣿⣿⣿⣿⡇⠀⢠⣿⣿⣿⣿⣷⡌⢻⣿⡇⢠⣿
⠀⠀⠀⠀⠛⠀⠀⠀⣿⣿⣿⣿⡀⢀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣇⠀⣴⣿⣿⣿⣿⣿⡟⢠⠉⠀⣼⣿
⠀⠀⠀⠀⣶⣄⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠟⠛⠋⠛⡛⠛⠛⢿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡟⠀⣿⣿⣶⣿⣿
⠀⠀⠀⣠⣿⡿⠃⣼⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⣿⣿⣿⣿⣿⡇⢸⣿⣿⣿⣿⣿⣿⣿⣿⠏⢠⣾⣿⣿⣿⣿⣿
⢀⣴⣾⣿⣿⡀⠸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣆⠘⣿⣿⣿⠿⢀⣿⣿⣿⣿⣿⣿⣿⣿⠏⢠⣾⣿⣿⣿⣿⣿⣿
⢼⠟⠛⢿⣿⣿⣤⡈⠛⠿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣧⣄⣉⣁⣴⣿⣿⣿⣿⣿⣿⣿⡿⠛⢀⣿⣿⡿⢿⣿⣿⣿⣿

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `TensorSerializer.write_module` has been optimized further for a speedup of
+  ~3.6x on CUDA modules and ~3.1x faster on CPU modules
+
 ### Fixed
 
-- `CURLStreamFile.response_headers` no longer has a chance to have incomplete
+- `CURLStreamFile.response_headers` no longer has a chance to contain incomplete
   header information
 
 ## [2.3.0] - 2023-09-06
--- a/tensorizer/_version.py
+++ b/tensorizer/_version.py
@@ -1,1 +1,1 @@
-__version__ = "2.3.1.dev0"
+__version__ = "2.3.1.dev1"
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -1818,7 +1818,7 @@ class TensorSerializer:
         # This task is I/O-bound and has no prerequisites,
         # so it goes into the regular writer pool.
         def write_metadata():
-            assert self._pwrite(metadata, metadata_pos)
+            self._pwrite(metadata, metadata_pos)
 
         self._jobs.append(self._writer_pool.submit(write_metadata))
 
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -1988,12 +1988,15 @@ class TensorSerializer:
         next_pos = self._file.tell()
 
         fallocate = getattr(os, "posix_fallocate", None)
-        if fallocate:
+        if fallocate and self._fd:
             size += sum(t.untyped_storage().size() for t in tensors)
             # Rough underestimate of header size
             header_min_size = 24
             size += header_min_size * len(tensors)
-            fallocate(self._file.fileno(), next_pos, size)
+            try:
+                fallocate(self._fd, next_pos, size)
+            except OSError:
+                pass
 
         cuda_tensors = [t for t in tensors if t.device.type == "cuda"]
         if cuda_tensors:
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -1315,7 +1315,7 @@ class TensorSerializer:
         # to each pool in the same relative order.
 
         # Tracks work submitted to all pools to wait for pending work to finish.
-        self._jobs = []
+        self._jobs: List[concurrent.futures.Future] = []
 
         if self.compress_tensors:
             import lz4.frame
@@ -1476,7 +1476,9 @@ class TensorSerializer:
         for pool in "_computation_pool", "_writer_pool", "_header_writer_pool":
             thread_pool = getattr(self, pool, None)
             if thread_pool is not None:
-                thread_pool.shutdown(wait=False, cancel_futures=True)
+                for j in self._jobs:
+                    j.cancel()
+                thread_pool.shutdown(wait=False)
 
     def _synchronize_pools(self):
         for j in self._jobs:
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -12,19 +12,35 @@ except ImportError:
     resource = None
 
 import collections.abc
+import concurrent.futures
 import ctypes
+import dataclasses
 import hashlib
 import io
 import itertools
 import logging
 import mmap
 import os
+import queue
 import struct
 import threading
 import time
 import typing
 import zlib
+from collections import OrderedDict
 from enum import Enum
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 import numpy
 import torch
@@ -37,9 +53,6 @@ if torch.cuda.is_available():
     cudart = torch.cuda.cudart()
 else:
     cudart = None
-
-from collections import OrderedDict
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 lz4 = None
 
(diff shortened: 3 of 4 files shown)


wbrown (approved):
Very good job. :+1: If we added some benchmarks in the `README.md` comparing this to `torch.save()`, this would drive home how fast this is.

PR #48, opened 2023-09-13
refactor: Reorganize Deserialization Parsing

Eta0:
# Deserialization Reorganization

This change groups encoding and decoding logic for header and metadata entries into dedicated classes, completely separate from the `TensorSerializer` and `TensorDeserializer` classes. `TensorSerializer` was already mostly updated to do this in #46, so this mainly affects `TensorDeserializer`.

This disentangles most of the *parsing* of `tensorizer`'s custom binary format from the *use* of that data, making it easier to update and optimize the deserializer's major bottlenecks like I/O, hash computation, and device transfer as was done in the serializer.


wbrown (approved):
Great job. This will make it much easier to work with. I like it better than the original code I wrote, with manual unpacks.

PR #49, opened 2023-09-14
(feat) add  `buffer_size ` argument to `CURLStreamFile`, adjust to `2mb`, add `benchmark.py`

wbrown:
This adds the `buffer_size` argument to `CURLStreamFile` to allow us to experiment with and adjust the `Popen` buffer. Getting the right buffer size appears to impact the speed of transfers. Too large a buffer means that it gets filled, and then the TCP socket exerts backpressure for a whle. Too small means that we're spending time waiting on the buffers to be repopulated.

In this spirit, `benchmark.py` was written to measure the impacts of read sizes and buffer sizes. Preliminary results imply that `2mb `is the most optimal at this time, so the default Popen buffer size has been set to `2mb`.


tensorizer/stream_io.py
```diff
@@ -142,6 +142,7 @@ def __init__(
         begin: Optional[int] = None,
         end: Optional[int] = None,
         headers: Dict[str, Any] = None,
+        buffer_size: int = 2**21,  # `2mb` buffer on the python IO object
```
Eta0, line 145: We can make this a keyword-only argument to make it easier to change later
```suggestion
        *,
        buffer_size: int = 2 << 20,  # 2 MiB buffer on the Python IO object
```

benchmark.py
Eta0, line 1: Please install the pre-commit hooks with `pip install pre-commit && pre-commit install` to auto-format code with `black` and `isort`, otherwise it makes the commit diff weirder for the next person to touch the file who does have them installed

tensorizer/stream_io.py
Eta0, line 1: Should update `tensorizer/_version.py` to `2.4.0.dev0` for the feature addition, and add a note about the new feature in the *Unreleased* section of the changelog for now

Eta0 (changes requested):

benchmark.py
```diff
@@ -0,0 +1,107 @@
...
+import torch
+import time
+from tensorizer.serialization import TensorDeserializer
+from tensorizer.stream_io import CURLStreamFile
+import os
+import sys
+
+# Read in model name from command line, or env var, or default
+# to gpt-j-6b
+model_name = os.getenv("MODEL_NAME", "EleutherAI/gpt-j-6b/fp16")
```
harubaru, line 11: Seems like the model name is case sensitive since when I tested this I ran into a 403 error.
```suggestion
model_name = os.getenv("MODEL_NAME", "EleutherAI/gpt-j-6B/fp16")
```

harubaru (changes requested):

PR #50, opened 2023-09-19
(feat)Add preliminary Redis support.

wbrown:
This is a preliminary implementation of Redis support by implementing a `RedisStreamFile` class.

```
    RedisStreamFile implements a file-like object around a Redis key namespace. Each
    'file' is broken up into multiple keys, each of which is a slice of the file. Each
    key is named with the prefix, followed by a colon, a user-assigned name, another
    colon, and then the byte index of the key.

    For example, if the prefix is 'foo', and the user-assigned name is 'bar', then the
    keys would be 'foo:bar:0', 'foo:bar:16'. The first key would be the first 16 bytes
    of the file, and the remainder would be in the next key.

    On initialization, it performs a prefix scan of the keys for the prefix, and then
    sorts the keys by their byte index. This allows it to perform a seek operation
    efficiently, as it can find the key that contains the byte index, and then read
    from that key.

    RedisStreamFile has some optimizations, such as reading the entirety of a key into
    a buffer, and then serving subsequent reads from that buffer. It also has a buffer
    for the remainder of a key, so that it can serve partial reads of a key.

    Arguments:
        uri: The URI of the Redis key namespace. This should be prefixed by the
            'redis://' scheme, and include the key prefix. For example,
            'redis://localhost:6379/foo'.
        buffer_size: The size of the TCP buffer to use for the Redis connection.

    Attributes:
        response_latencies: A list of the time it took to get the Redis responses.
        bytes_read: The number of bytes read from the stream.
        bytes_skipped: The number of bytes skipped from the stream.
        read_operations: The number of read operations performed on the stream.
```

The `TensorizerDeserializer` object has a new method, `to_redis()` that will populate a Redis server with tensors. For each file, it:
* Creates a key for the file metadata
... then it iterates over the tensors, creating two keys:
* Tensor metadata
* Tensor data

It's pretty quick, but this Redis support is still preliminary as it hasn't shown itself to give enough value vs just using S3. The trouble is going to be distributing enough instances and being semi-rack local.

The use cases I had in mind for this was for:
* Loading LoRAs on a per-request basis for inference
* kv cache between inference instances

```
2023-10-03 15:17:50 INFO     g32f8f4 -- curl:  gpu: NVIDIA RTX A6000 (47 GiB), raw read 5430.50 MiB at 1001.08 MiB/s, 8192.0 KiB stream buffer size, 256.0 KiB read size, cached: HIT by lga1-7d9984f84d-ntg55
2023-10-03 15:17:54 INFO     g32f8f4 -- redis: gpu: NVIDIA RTX A6000 (47 GiB), raw read 5430.66 MiB at 1594.09 MiB/s, 8192.0 KiB stream buffer size
2023-10-03 15:17:57 INFO     g32f8f4 -- redis: gpu: NVIDIA RTX A6000 (47 GiB), loaded   5430.66 MiB at 1532.85 MiB/s, 8192.0 KiB stream buffer size, plaid: True, verify_hash: False, lazy_load: True
2023-10-03 15:18:03 INFO     g32f8f4 -- redis: gpu: NVIDIA RTX A6000 (47 GiB), loaded   5430.66 MiB at 1109.77 MiB/s, 8192.0 KiB stream buffer size, plaid: False, verify_hash: False, lazy_load: True
2023-10-03 15:18:09 INFO     g32f8f4 -- curl:  gpu: NVIDIA RTX A6000 (47 GiB), loaded   5430.66 MiB at 848.45 MiB/s, 8192.0 KiB stream buffer size, plaid: True, verify_hash: False, lazy_load: True, cached: HIT by lga1-7d9984f84d-cghv5
2023-10-03 15:18:18 INFO     g32f8f4 -- curl:  gpu: NVIDIA RTX A6000 (47 GiB), loaded   5430.66 MiB at 632.85 MiB/s, 8192.0 KiB stream buffer size, plaid: False, verify_hash: False, lazy_load: True, cached: HIT by lga1-7d9984f84d-ntg55
```

~~**NOTE:** I am not sure why the test cases are failing on GHA, it appears that it can't reach `localhost`?~~ Solved.

### DIFF
--- a/examples/benchmark_buffer_size/benchmark.py
+++ b/examples/benchmark_buffer_size/benchmark.py
@@ -245,7 +245,7 @@ def bench_redis(
 
     del test_dict
 
-    if hasattr(torch, "cuda"):
+    if hasattr(torch, "cuda") and torch.cuda.is_available():
         torch.cuda.synchronize()
     gc.collect()
 
--- a/examples/benchmark_buffer_size/benchmark.py
+++ b/examples/benchmark_buffer_size/benchmark.py
@@ -41,7 +41,7 @@ http_uri = (
     f"/{model_name}/model.tensors"
 )
 
-print(f"Testing {http_uri}")
+logging.info(f"Testing {http_uri}")
 
 kibibyte = 1 << 10
 mebibyte = 1 << 20
@@ -55,9 +55,11 @@ try:
     cudadev = torch.cuda.current_device()
     gpu_gb = int(torch.cuda.get_device_properties(0).total_memory / gibibyte)
     gpu_name = torch.cuda.get_device_name(cudadev)
+    has_gpu = True
 except AssertionError:
     gpu_gb = 0
     gpu_name = "CPU"
+    has_gpu = False
 
 
 redis_client = redis.Redis(host="localhost", port=6379, db=0)
@@ -88,8 +90,8 @@ def io_test(
 
     # Print the total size of the stream, and the speed at which it was read.
     logging.info(
-        f"{nodename} -- "
-        f"gpu: {gpu_name} ({gpu_gb} GiB), streamed "
+        f"{nodename} -- curl:  "
+        f"gpu: {gpu_name} ({gpu_gb} GiB), raw read "
         f"{total_sz / mebibyte:0.2f} MiB at "
         f"{total_sz / mebibyte / (end - start):0.2f} MiB/s, "
         f"{buffer_size / kibibyte} KiB stream buffer size, "
@@ -125,7 +127,7 @@ def deserialize_test(
     total_sz = test_dict.total_bytes_read
 
     logging.info(
-        f"{nodename} -- "
+        f"{nodename} -- curl:  "
         f"gpu: {gpu_name} ({gpu_gb} GiB), loaded  "
         f" {total_sz / mebibyte:0.2f} MiB at"
         f" {total_sz / mebibyte / (end - start):0.2f} MiB/s,"
@@ -160,32 +162,48 @@ def load_redis():
     test_dict.to_redis(redis_client, model_name)
     end = time.monotonic()
     rate = test_dict.total_bytes_read / (end - start)
-    print(
-        f"Loaded {test_dict.total_bytes_read / gibibyte:0.2f} GiB at"
+    logging.info(
+        f"redis: Loaded {test_dict.total_bytes_read / gibibyte:0.2f} GiB at"
         f" {rate / mebibyte:0.2f} MiB/s in {end - start:0.2f}s"
     )
 
 
-def bench_redis(buffer_size=256 * kibibyte):
+def bench_redis(
+    plaid_mode=False,
+    verify_hash=False,
+    lazy_load=False,
+    buffer_size=256 * kibibyte,
+):
     test_dict = TensorDeserializer(
         RedisStreamFile(
             f"redis://localhost:6379/{model_name}", buffer_size=buffer_size
         ),
-        lazy_load=True,
-        plaid_mode=True,
+        lazy_load=lazy_load,
+        plaid_mode=plaid_mode,
+        verify_hash=verify_hash,
     )
 
-    all_begin = time.monotonic()
+    start = time.monotonic()
     for name in test_dict:
         test_dict[name]
-    all_end = time.monotonic()
+    end = time.monotonic()
+    total_sz = test_dict.total_bytes_read
 
-    print(
-        f"Read {test_dict.total_bytes_read / gibibyte:0.2f} GiB at"
-        f" {buffer_size} bytes stream buffer size in"
-        f" {test_dict.total_bytes_read / gibibyte / (all_end - all_begin):0.2f} GiB/s"
+    logging.info(
+        f"{nodename} -- redis: "
+        f"gpu: {gpu_name} ({gpu_gb} GiB), loaded  "
+        f" {total_sz / mebibyte:0.2f} MiB at"
+        f" {total_sz / mebibyte / (end - start):0.2f} MiB/s,"
+        f" {buffer_size / kibibyte} KiB stream buffer size, plaid:"
+        f" {plaid_mode}, verify_hash: {verify_hash}, lazy_load:"
+        f" {lazy_load or plaid_mode}"
     )
+
     del test_dict
+
+    if hasattr(torch, "cuda"):
+        torch.cuda.synchronize()
+    gc.collect()
 
 
 def io_test_redis(buffer_size=256 * kibibyte):
@@ -197,9 +215,9 @@ def io_test_redis(buffer_size=256 * kibibyte):
 
     buf = bytearray(512 * mebibyte)
 
-    total_bytes = 0
+    total_sz = 0
 
-    start_time = time.monotonic()
+    start = time.monotonic()
 
     # Loop over redis keys, and read them into memory.
     for key in redis_client.scan_iter(f"{model_name}:*"):
@@ -223,16 +241,18 @@ def io_test_redis(buffer_size=256 * kibibyte):
             num_bytes = redis_tcp.recv_into(mv, left, socket.MSG_WAITALL)
             mv = mv[num_bytes:]
             left -= num_bytes
-            total_bytes += num_bytes
+            total_sz += num_bytes
             read_ct += 1
         redis_tcp.recv(2)
 
-    end_time = time.monotonic()
+    end = time.monotonic()
     redis_tcp.close()
-    print(
-        f"Raw read {total_bytes / gibibyte:0.2f} GiB at"
-        f" {buffer_size} bytes stream buffer size in"
-        f" {total_bytes / gibibyte / (end_time - start_time):0.2f} GiB/s"
+    logging.info(
+        f"{nodename} -- redis: "
+        f"gpu: {gpu_name} ({gpu_gb} GiB), raw read "
+        f"{total_sz / mebibyte:0.2f} MiB at "
+        f"{total_sz / mebibyte / (end - start):0.2f} MiB/s, "
+        f"{buffer_size / kibibyte} KiB stream buffer size"
     )
 
 
@@ -242,8 +262,12 @@ for buffer_size_power in range(16, 21):
     for sample in range(5):
         io_test(buffer_size=buffer_size)
         io_test_redis(buffer_size=buffer_size)
-        bench_redis(buffer_size=buffer_size)
-        deserialize_test(buffer_size=buffer_size)
+        if has_gpu:
+            bench_redis(buffer_size=buffer_size, plaid_mode=True)
+        bench_redis(buffer_size=buffer_size, lazy_load=True)
+        if has_gpu:
+            deserialize_test(buffer_size=buffer_size, plaid_mode=True)
+        deserialize_test(buffer_size=buffer_size, lazy_load=True)
 
 
 exit(0)
--- a/examples/benchmark_buffer_size/benchmark.py
+++ b/examples/benchmark_buffer_size/benchmark.py
@@ -141,12 +141,18 @@ def deserialize_test(
     plaid_mode=False,
     verify_hash=False,
     lazy_load=False,
+    force_http=False,
     buffer_size=256 * kibibyte,
 ):
     scheme = source.split("://")[0]
+    if scheme == "s3" and not force_http:
+        scheme = "s3s"
     scheme_pad = " " * (5 - len(scheme))
     source = open_stream(
-        source, s3_endpoint=default_s3_read_endpoint, buffer_size=buffer_size
+        source,
+        s3_endpoint=default_s3_read_endpoint,
+        buffer_size=buffer_size,
+        force_http=force_http,
     )
     start = time.monotonic()
     test_dict = TensorDeserializer(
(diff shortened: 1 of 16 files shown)


tensorizer/serialization.py
```diff
@@ -1488,6 +1493,34 @@ def verify_module(
...
         return all(result for name, result in results), results
 
+    def to_redis(
+        self, redis_client: redis.Redis, key_prefix: str, force: bool = False
+    ) -> None:
+        """
+        Given a redis client and a key_prefix, write the tensors in this
+        Tensorizer object to the redis client under the given key prefixes.
+        """
+        header_entry = b"|TZR|"
```
harubaru, line 1503: Might be good to have a try-catch block here in case the client spews out an error for easier debugging.

harubaru (approved):
I've ran through the Redis benchmarks without any issues. Though it may be useful to add a README to this benchmark to add more information on its purpose and how to set it up.

tests/test_serialization.py
```diff
@@ -337,6 +338,47 @@ def test_s3_lazy_load(self):
...
+                device=default_device,
+                lazy_load=True,
+            )
+            check_deserialized(self, deserialized_redis, model_name)
+            check_inference(
+                self, deserialized_redis, model_name, default_device
+            )
+            deserialized_redis.close()
+
+        test_redis.tearDown()
```
Eta0, line 380: This would be good to put into a `finally:` clause in case any of the preceding subtests fail

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+    prefix = uri_components.path.lstrip("/")
+
+    return host, port, prefix
+
+
+# Detect if we're running on OSX, and if so, set max buffer size to 1 MiB.
+if sys.platform == "darwin":
+    _MAX_TCP_BUFFER_SIZE = 2 << 20  # 1 MiB if OSX
+else:
+    _MAX_TCP_BUFFER_SIZE = 2 << 22  # 8 MiB
```
Eta0, line 444: The "1 MiB" case was actually 2 MiB. The other change is for readability.
```suggestion
if sys.platform == "darwin":
    _MAX_TCP_BUFFER_SIZE = 1 << 20  # 1 MiB if OSX
else:
    _MAX_TCP_BUFFER_SIZE = 8 << 20  # 8 MiB
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+    """
+
+    def __init__(
+        self,
+        uri: str,
+        *,
+        buffer_size: int = _MAX_TCP_BUFFER_SIZE,
+    ) -> None:
+        if buffer_size is None:
+            buffer_size = _MAX_TCP_BUFFER_SIZE
```
Eta0, line 488: Same as with cURL: `buffer_size` is not typed as an `Optional[int]`, so this is an illegal value.
```suggestion
        if buffer_size is None:
            raise TypeError("Invalid buffer_size")
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+        init_begin = time.monotonic()
+        host, port, prefix = _parse_redis_uri(uri)
+        self._redis = redis.Redis(host=host, port=port, db=0)
+        self._redis_tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self._redis_tcp.setsockopt(
+            socket.SOL_SOCKET, socket.SO_RCVBUF, buffer_size
+        )
+        self._redis_tcp.connect((host, port))
+        self._redis_tcp_mutex = threading.Lock()
```
Eta0, line 497: Multithreading is not used in this file's own code; what is the justification for guarding `_read_from` with a mutex? Concurrent `read` calls on file objects are usually already considered unsafe and protected by locks in the caller's code if they desire concurrent access, especially in our own `serialization.py`.

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+        # Do a key scan for the prefix, and collect all the indexes from the keys.
+        self._keys = []
+        self._indexes = []
+        self._sizes = []
+        keys = self._redis.scan_iter(f"{prefix}:*")
+        # Sort the keys by their index.
+        for key in sorted(
+            keys, key=lambda x: int(x.decode("utf-8").split(":")[-1])
+        ):
+            self._keys.append(key)
```
Eta0, line 508: `sorted()` returns a `list`.
```suggestion
        self._indexes = []
        self._sizes = []
        keys = self._redis.scan_iter(f"{prefix}:*")
        # Sort the keys by their index.
        self._keys = sorted(
            keys, key=lambda x: int(x.decode("utf-8").split(":")[-1])
        )
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+            < position_end
+            < curr_buffer_end
+        ):
+            num_bytes = min(read_sz, len(self._curr_buffer_view))
+            ba[:num_bytes] = self._curr_buffer_view[:num_bytes]
+            self._curr_buffer_idx += num_bytes
+            self._curr_buffer_view = self._curr_buffer_view[num_bytes:]
+            return num_bytes
+
+        self._redis_tcp_mutex.acquire(True)
```
Eta0, line 572: I'd suggest dropping the mutex in this code—but if it's kept, it should use the `with self._redis_tcp_mutex:` form for more exception safety

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+            if tcp_read_sz > len(self._curr_buffer):
+                self._curr_buffer = bytearray(tcp_read_sz)
+            self._curr_buffer_view = memoryview(self._curr_buffer)[:tcp_read_sz]
+            self._redis_tcp.recv_into(
+                self._curr_buffer_view, tcp_read_sz, socket.MSG_WAITALL
+            )
+            self._curr_buffer_idx = position + num_bytes
+
+        # Read the trailing \r\n and discard.
+        self._redis_tcp.recv(2)
```
Eta0, line 610: Extra sanity check for good measure
```suggestion
        if self._redis_tcp.recv(2) != b"\r\n":
            raise RuntimeError("Missing key footer")
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+
+        self._redis_tcp_mutex.release()
+        return num_bytes
+
+    def _read_sz(self) -> int:
+        # Loop until we get a \r\n
+        sz_resp = b""
```
Eta0, line 617: Time complexity, as mentioned in another comment
```suggestion
        sz_resp = bytearray()
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+        while True:
+            b = self._redis_tcp.recv(1)
+            sz_resp += b
+            if sz_resp[-2:] == b"\r\n":
+                break
```
Eta0, line 622: Prevents an infinite loop if the connection is interrupted (assuming the socket isn't non-blocking)
```suggestion
        while True:
            b = self._redis_tcp.recv(1)
            if b == b"":
                raise IOError("Failed to read size")
            sz_resp += b
            if sz_resp[-2:] == b"\r\n":
                break
```

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+                if num_bytes == 0:
+                    break
+                left -= num_bytes
+                self._curr += num_bytes
+                if left == 0:
+                    break
+                mv = mv[num_bytes:]
+            self.bytes_read += rq_sz - left
+            if ba is None:
+                return orig_mv.tobytes()
```
Eta0, line 651: This requires an extra `memcpy` of all of the data from the mutable `bytearray` to the new immutable `bytes` object. If it is permissible to return a `bytearray` from this method, returning `orig_mv.obj` instead would be more performant.

tensorizer/stream_io.py
```diff
@@ -417,6 +422,292 @@ def seek(self, position, whence=SEEK_SET):
...
+            else:
+                return rq_sz - left
+        except (IOError, OSError) as e:
+            raise e
```
Eta0, line 655: Redundant
```suggestion
```

Eta0 (changes requested):

examples/requirements.txt
```diff
@@ -1,4 +1,5 @@
 transformers~=4.27.1
 diffusers==0.14.0
 scipy~=1.10.0
-accelerate==0.17.1
\ No newline at end of file
+accelerate==0.17.1
+redis
```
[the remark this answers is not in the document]
Eta0: Actually, if `redis` is added to `pyproject.toml` (which it needs to be), this should just be deleted from `examples/requirements.txt`, since this only lists supplementary requirements beyond those needed for the package.
```suggestion
```

examples/benchmark_buffer_size/benchmark.py
```diff
@@ -125,17 +179,150 @@ def deserialize_test(
 
     test_dict.close()
     del test_dict
-    torch.cuda.synchronize()
+    if hasattr(torch, "cuda"):
```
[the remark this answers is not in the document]
wbrown: Hmm, good one!

examples/benchmark_buffer_size/benchmark.py
```diff
@@ -125,17 +179,150 @@ def deserialize_test(
...
+            deserialize_test(
+                source=https_uri, buffer_size=buffer_size, plaid_mode=True
+            )
+        deserialize_test(source=s3_uri, buffer_size=buffer_size, lazy_load=True)
+        if has_gpu:
+            deserialize_test(
+                source=s3_uri, buffer_size=buffer_size, plaid_mode=True
+            )
+
+exit(0)
```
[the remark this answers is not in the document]
wbrown: Ha, that was an `exit(0)` for an early exit before debugging -- snuck down to the bottom.

requirements.txt
```diff
@@ -2,4 +2,6 @@ torch>=1.9.0
 numpy>=1.19.5
 protobuf>=3.19.5
 psutil>=5.9.4
-boto3>=1.26.0
\ No newline at end of file
+boto3>=1.26.0
+hiredis
```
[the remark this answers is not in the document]
wbrown: This actually is used. It's one of these 'if this exists, `redis` will use it.

tensorizer/stream_io.py
```diff
@@ -145,6 +148,8 @@ def __init__(
         *,
         buffer_size: int = 2 << 20,  # 2 MiB buffer on the Python IO object
     ) -> None:
+        if buffer_size is None:
+            buffer_size = 2 << 20
```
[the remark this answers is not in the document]
wbrown: This is needed for the top level call to pass in `None`.

tensorizer/stream_io.py
wbrown: As mentioned earlier, `None` will get passed in as a top level value.

tests/test_serialization.py
```diff
@@ -28,6 +28,7 @@
     utils,
 )
 from tensorizer.serialization import TensorHash, TensorType
+from test_stream_io import TestRedis
```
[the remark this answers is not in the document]
wbrown: Fixed, not into a supporting file for now though.

tensorizer/stream_io.py
wbrown: Didn't know about `obj`, as the type was opaque to me on inspection. Cool!

tensorizer/serialization.py
```diff
@@ -1488,6 +1493,34 @@ def verify_module(
 
         return all(result for name, result in results), results
 
+    def to_redis(
+        self, redis_client: redis.Redis, key_prefix: str, force: bool = False
+    ) -> None:
+        """
+        Given a redis client and a key_prefix, write the tensors in this
+        Tensorizer object to the redis client under the given key prefixes.
+        """
```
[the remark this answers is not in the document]
wbrown: Expanded documentation.

requirements.txt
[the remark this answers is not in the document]
wbrown: Done, along with `hiredis`.

examples/benchmark_buffer_size/benchmark.py
```diff
@@ -125,17 +179,150 @@ def deserialize_test(
...
+        ),
+        lazy_load=lazy_load,
+        plaid_mode=plaid_mode,
+        verify_hash=verify_hash,
+    )
+
+    start = time.monotonic()
+    for name in test_dict:
+        test_dict[name]
+    end = time.monotonic()
```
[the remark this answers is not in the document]
wbrown: 1 is addressed. We'll address 2 when it's time.

tests/test_serialization.py
```diff
@@ -337,6 +338,47 @@ def test_s3_lazy_load(self):
...
+            f"redis://{test_redis.redis_host}:{test_redis.redis_port}"
+            f"/{model_name}"
+        )
+
+        deserialized_s3 = TensorDeserializer(
+            f"s3://tensorized/{model_name}/model.tensors",
+            device=default_device,
+        )
+        deserialized_s3.to_redis(test_redis.redis_client, model_name)
+        deserialized_s3.close()
```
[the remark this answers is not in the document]
wbrown: Addressed. :+1:

tensorizer/stream_io.py
wbrown: Internal for overlapping buffering and I/O. External reads do not mean internal reads.

tensorizer/stream_io.py
wbrown: Addressed.

PR #52, opened 2023-10-06
perf(serialization): Optimize `write_state_dict`

Eta0:
# Fix serialization with Python buffered I/O objects & optimize `write_state_dict`

## Stale Buffers
This change fixes a bug with serialization that manifested in files that were serialized directly to Kubernetes PVCs becoming corrupted (#55).
The issue was that `os.pwrite` bypasses Python-level write buffering on file-like objects used for serialization, causing any write buffers to become stale and potentially overwrite good data when flushed *after* later `os.pwrite` calls.
The fix in this PR calls `flush()` (if available on the file-like object) after the two small sequential write sequences during serialization to never leave data sitting uncommitted in a write buffer between serializer methods.

A workaround on earlier versions of tensorizer is to use `buffering=0` on the file-like object used for serialization to disable write buffers. This fix is necessary to avoid corrupting files for people who aren't aware of the workaround, and to potentially improve performance on the sequential write portion.

## `write_state_dict`
The other included change generalizes the optimizations to `TensorSerializer.write_module` from #46, and applies them to `TensorSerializer.write_state_dict` as well, speeding that method up by a similar amount (about 3x).


wbrown:
@Eta0 For any and all 'performance' PRs, we really need to show what changed performance wise. We have a nice performance collection framework now.

harubaru (approved):

PR #54, opened 2023-10-09
`plaid_mode_buffers` argument, JSON support, `lazy_load` assumptions

wbrown:
* Adds `plaid_mode_buffers` argument to `TensorDeserializer` that (currently) defaults to `4`
* Adds JSONL output support to benchmarking, DRY's up logging
* Adds `--start_plaid_buffers`, `--end_plaid_buffers`, `--iterations`, `--test-s3`, `--test-https` arguments to `benchmark.py`.

### DIFF
--- a/examples/benchmark_buffer_size/benchmark.py
+++ b/examples/benchmark_buffer_size/benchmark.py
@@ -57,16 +57,40 @@ parser.add_argument(
     help="Don't load the model into redis",
 )
 parser.add_argument(
-    "--start",
+    "--start_buffer_size",
     type=int,
     default=18,
     help="Starting buffer size power (default: 18)",
 )
 parser.add_argument(
-    "--end",
+    "--end_buffer_size",
     type=int,
     default=28,
     help="Ending buffer size power (default: 28)",
+)
+parser.add_argument(
+    "--start_plaid_buffers",
+    type=int,
+    default=1,
+    help="Starting plaid buffers power (default: 1)",
+)
+parser.add_argument(
+    "--end_plaid_buffers",
+    type=int,
+    default=4,
+    help="Ending plaid buffers power (default: 4)",
+)
+parser.add_argument(
+    "--test-https",
+    action="store_true",
+    default=False,
+    help="Test HTTPS download speeds",
+)
+parser.add_argument(
+    "--test-s3",
+    action="store_true",
+    default=False,
+    help="Test S3 download speeds",
 )
 parser.add_argument(
     "--json",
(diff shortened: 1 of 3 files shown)


examples/benchmark_buffer_size/benchmark.py
```diff
@@ -50,17 +57,53 @@
...
+    type=int,
+    default=1,
+    help="Starting plaid buffers power (default: 1)",
+)
+parser.add_argument(
+    "--end_plaid_buffers",
+    type=int,
+    default=4,
+    help="Ending plaid buffers power (default: 4)",
+)
```
Eta0, line 82: Should have dashed names rather than underscores for consistency

examples/benchmark_buffer_size/benchmark.py
```diff
@@ -90,16 +147,129 @@
...
+plaid_sizes = [
+    1 << i
+    for i in range(args.start_plaid_buffers - 1, args.end_plaid_buffers + 1)
+]
+
+
+def log(
+    total_sz: int,
+    start: float,
+    end: float,
```
Eta0, line 174: Individual `time.monotonic()` timepoints are defined as have no meaning—they aren't actual "date" timestamps and can't be converted to absolute dates or reliably compared with other systems—only their differences have meaning, so it would be more helpful to just log `end - start` durations in these

tensorizer/serialization.py
```diff
@@ -690,6 +693,7 @@ def __init__(
         *,
         lazy_load: bool = False,
         plaid_mode: bool = False,
+        plaid_mode_buffers: int = 4,
```
Eta0, line 696: Should make this a `ValueError` to specify when `plaid_mode` is `False` to prevent the predictable mistake of someone specifying only `plaid_mode_buffers` and expecting that to imply `plaid_mode=True`
```suggestion
        plaid_mode_buffers: Optional[int] = None,
```

tensorizer/serialization.py
```diff
@@ -781,7 +785,7 @@ def __init__(
                 for name, entry in self._metadata.items()
             }
             self.total_tensor_bytes = sum(tensor_sizes.values())
-            self._plaid_mode_buffer_count = 4
+            self._plaid_mode_buffer_count = plaid_mode_buffers
```
Eta0, line 788: Continuation of a previous suggestion:
```suggestion
            self._plaid_mode_buffer_count = plaid_mode_buffers or 4
            if not self._plaid_mode and plaid_mode_buffers is not None:
                raise ValueError("Cannot specify plaid_mode_buffers when plaid_mode=False")
```

Eta0 (changes requested):

Eta0 (approved):

PR #56, opened 2023-10-16
fix: Improve error tracebacks from `TensorDeserializer._bulk_load`

Eta0:
# Improve error tracebacks from `TensorDeserializer._bulk_load`

Unconditionally cancelling worker threads was causing their traceback printouts to be interrupted by a secondary, injected exception. This change makes the main thread wait for a short delay after an exception in a worker thread, to allow for graceful shutdown before forcing another exception upon it.

This change also switches the `Cancelled` exception to derive from `BaseException` to make it more consistent at cancelling worker threads, and silences traceback printouts from `Cancelled` exceptions, since they are secondary to the original exceptions and provide little useful info.


harubaru (approved):
Works in providing the real error :+1:

PR #57, opened 2023-10-19
feat(serialization): Support serializing only persistent buffers

Eta0:
# Non-Persistent Buffer Serialization

As previously noted in #36, calling `TensorSerializer.write_module` serializes non-persistent buffers in addition to persistent buffers—in contrast to the behaviour of serializing a `state_dict`, which excludes non-persistent buffers. This change introduces a new `include_non_persistent_buffers` keyword-only boolean parameter for `write_module` that can toggle off the serialization of non-persistent buffers.

This defaults to `True` to match the old behaviour so as not to be a breaking change for anyone, but the default may be swapped in a later version (e.g. 3.0.0) of `tensorizer`.

The corresponding changelog entry follows:

> - `TensorSerializer.write_module` now accepts `include_non_persistent_buffers` as a keyword-only boolean argument that can be set to `False` to exclude buffers from serialization that were originally registered to the module through calling `torch.nn.Module.register_buffer` with `persistent=False`
>   - `torch.nn.Module.state_dict` never includes persistent buffers, so setting this to `False` will more closely match the behaviour of `state_dict` serialization
>   - `TensorSerializer.write_module` used to always include non-persistent buffers
>   - The default (`include_non_persistent_buffers=True`) matches the old behaviour


wbrown (approved):
Approved, but I'm not sure if it's desired behavior to serialize non-persistent buffers, even if it's technically a 'breaking change'.

I think we should do the more correct default behavior here.

wbrown:
Eta. from Slack:
> my concern is that defaulting to not serializing some tensors will mean that people using the no_init_or_tensor idiom will be left with some tensors on the meta device that would have been initialized as a side-effect of tensorizer before
> I.e. they'd never have memory allocated for non-persistent buffers, beyond just not getting their contents
> since we haven't tested whether it's safe to swap out in that case I'd rather just make the default the same as before for now, since we can just toggle it to False if we rule that out as an issue

On discussion, agreed, and merging.

PR #73, opened 2024-01-31
feat(examples): Update `hf_serialization.py` to be used as Kubernetes Job container

sangstar:
# Updating `hf_serialization.py` for use in Kubernetes Jobs that can serialize models for inference services
This change allows containerization of `hf_serialization` to be used in Kubernetes Jobs that can run before an inference service to ensure a model that is meant to be deserialized by pods is serialized. S3 authorization can now be performed in the absence of seeking a `.s3cfg` by providing `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables. 

Model config file uploads and downloads are also supported in order to completely avoid intermediating with HuggingFace where possible. `--force` has been added as a CLI command that writes model tensors and config even if files exist at the desired write location. Otherwise, the file will be checked by reading the first byte, and if successful, will skip serialization, additionally allowing `hf_serialization.py` to be run idempotently. 

Model validation is also now done by ensuring model weights sameness rather than checking if it can generate text.


examples/hf_serialization.py
```diff
@@ -28,6 +28,28 @@
...
 
+s3_access_key_id = os.environ.get("AWS_ACCESS_KEY_ID")
+s3_secret_access_key = os.environ.get("AWS_SECRET_ACCESS_KEY")
+
+if os.environ.get("AWS_ENDPOINT_URL"):
+    default_s3_read_endpoint = os.environ.get("AWS_ENDPOINT_URL")
+    default_s3_write_endpoint = os.environ.get("AWS_ENDPOINT_URL")
+else:
+    default_s3_read_endpoint = "accel-object.ord1.coreweave.com"
+    default_s3_write_endpoint = "object.ord1.coreweave.com"
```
Eta0, line 39: 1. We aren't AWS
2. Use `or None` to map blank/empty (but still set) environment variables to `None` instead of `""`, since those should be treated equivalently (but `""` has special meaning as an access key to `open_stream`)
3. Setting the endpoint to `None` should already default to the `*object.ord1.coreweave.com` endpoints (or `.s3cfg` if available) through the logic in `open_stream`
```suggestion
s3_access_key_id = os.environ.get("S3_ACCESS_KEY_ID") or None
s3_secret_access_key = os.environ.get("S3_SECRET_ACCESS_KEY") or None
default_s3_write_endpoint  = default_s3_read_endpoint = (
    os.environ.get("S3_ENDPOINT_URL") or None
)
```

examples/hf_serialization.py
```diff
@@ -39,15 +61,43 @@
...
+        try:
+            with stream_io.open_stream(
+                file,
+                "rb",
+                *s3_read_credentials,
+            ) as f:
+                f.read(1)
+                return True
+        except OSError as e:
+            return False
```
Eta0, line 88: Tensorizer should technically return an empty string when reading from a file that is extant-yet-empty to behave like a file-like object. It doesn't at the moment—instead it raises an `OSError`—but if the implementation changes or this somehow manages to open a local file, `bool(f.read(1))` (equivalent to `len(f.read(1)) > 0`) would technically be more correct.
Also the error identifier `e` isn't used in the `except` clause below, so it doesn't need to be captured.
```suggestion
                return bool(f.read(1))
        except OSError:
            return False
```

examples/hf_serialization.py
```diff
@@ -39,15 +61,43 @@
...
+
+
 def serialize_model(
     model: torch.nn.Module,
     config: Optional[Union[ConfigMixin, AutoConfig, dict]],
     model_directory: str,
-    model_prefix: str = "model",
+    model_prefix: str = None,
```
Eta0, line 95: If the default is actually `"model"`, then it should be set as a default in the signature for clarity (rather than checking against `None` only to set it to `"model"` later)
```suggestion
    model_prefix: str = "model",
```

examples/hf_serialization.py
```diff
@@ -60,23 +110,40 @@ def serialize_model(
...
             serialized to the same directory. A good example are Stable
             Diffusion models. Default is "model".
+        force: Force upload serialized tensors to `output_prefix`
+            even if they already exist
     """
 
-    os.makedirs(model_directory, exist_ok=True)
-    dir_prefix = f"{model_directory}/{model_prefix}"
+    if model_prefix is None:
+        model_prefix = "model"
```
Eta0, line 119: (Continuing from the previous suggestion)
```suggestion
```

examples/hf_serialization.py
```diff
@@ -60,23 +110,40 @@ def serialize_model(
...
+        config_path = f"{dir_prefix}-config.json"
+        if (not config_file_exists) or force:
+            logger.info(f"Writing config to {config_path}")
+            with stream_io.open_stream(
+                config_path, "wb+", *s3_write_credentials
+            ) as f:
+                if hasattr(config, "to_dict"):
+                    f.write(bytes(json.dumps(config.to_dict()), "utf-8"))
+                elif isinstance(config, dict):
+                    f.write(bytes(json.dumps(config), "utf-8"))
```
Eta0, line 137: Slightly DRY-er
```suggestion
                config_dict = (
                    config.to_dict() if hasattr(config, "to_dict") else config
                )
                f.write(json.dumps(config_dict).encode("utf-8"))
```

examples/hf_serialization.py
```diff
@@ -226,6 +302,13 @@ def df_main(args: argparse.Namespace) -> None:
 
 def hf_main(args):
     output_prefix = args.output_prefix
+
+    # May not be necessary if users can be assumed won't accidentally
+    # include trailing slashes in their output_prefix
+    output_prefix = (
+        output_prefix[:-1] if output_prefix[-1] == "/" else output_prefix
+    )
```
Eta0, line 310:
```suggestion
    output_prefix = output_prefix.rstrip("/")
```

examples/hf_serialization.py
```diff
@@ -242,37 +325,36 @@ def hf_main(args):
...
-
-    tokenizer = AutoTokenizer.from_pretrained(args.input_directory)
-    tokenizer.save_pretrained(output_prefix)
+    serialize_model(model, model_config, output_prefix, None, args.force)
 
     if args.validate:
-        del model, model_config
+        # Not sure if this part is needed as, although I doubt it,
+        # I'm not absolutely certain whether or not TensorSerializer
+        # changes the model's state in any way
```
Eta0, line 333: `TensorSerializer` does not change the model's state (unless you pass a special flag to delete tensors after writing them, or an error occurs in the middle of model encryption). It used to by default a long time ago, which might be why that is there.

examples/hf_serialization.py
```diff
@@ -242,37 +325,36 @@ def hf_main(args):
     logger.info("GPU: " + utils.get_gpu_name())
     logger.info("PYTHON USED RAM: " + utils.get_mem_usage())
 
-    serialize_model(model, model_config, output_prefix)
-
-    tokenizer = AutoTokenizer.from_pretrained(args.input_directory)
```
wbrown, line 247: Why did we remove this? We do want to copy the tokenizer data, and this also updates the tokenizer to what format   is supported.

examples/hf_serialization.py
```diff
@@ -60,23 +110,40 @@ def serialize_model(
...
+    dir_prefix = f"{model_directory}/{model_prefix}"
+    config_file_exists, weights_file_exists = (
+        check_file_exists(f"{dir_prefix}-config.json"),
+        check_file_exists(f"{dir_prefix}.tensors"),
+    )
     if config is None:
         config = model
     if config is not None:
-        if hasattr(config, "to_json_file"):
-            config.to_json_file(f"{dir_prefix}-config.json")
```
wbrown, line 72: Why remove the propagation of the configuration file?

wbrown (changes requested):

examples/hf_serialization.py
```diff
@@ -242,37 +325,36 @@ def hf_main(args):
...
-            )
-        logger.info(
-            "Test Output:"
-            f" {tokenizer.decode(output[0], skip_special_tokens=True)}"
-        )
+        # Comparing model parameters
+        logger.info("Testing for sameness of model parameters")
+        for name, param in model.named_parameters():
+            param2 = model2.state_dict()[name]
+            assert torch.allclose(param, param2, atol=1e-3)
```
Eta0, line 357: 1. This check's logic will miss if the serialized model somehow got extra fields (like, say, duplicates) that weren't present in the original
2. That `logger.info` could probably be `logger.debug`
3. Calling `.state_dict()` is slow, so it's better to put that before the loop
4. The values should be exactly equal. If they slightly off enough to need `torch.allclose` rather than `torch.equal`, that's bad
5. `assert` isn't good for testing correctness because `assert` statements can be skipped
```suggestion
        tensorizer_model = load_model(
            output_prefix,
            AutoModelForCausalLM,
            AutoConfig,
            None,
            device,
            dtype,
        ).eval()
        # Comparing model parameters
        logger.debug("Testing for sameness of model parameters")
        original_sd = model.state_dict()
        tensorizer_sd = tensorizer_model.state_dict()
        if not tensorizer_sd.keys() == original_sd.keys():
            raise RuntimeError(
                "Validating deserialized model failed: different keys"
            )
        for name, original_tensor in original_sd.items():
            tensorizer_tensor = tensorizer_sd[name]
            if not torch.equal(original_tensor, tensorizer_tensor):
                raise RuntimeError(
                    "Validating deserialized model failed:"
                    f" weights for tensor {name} don't match"
                )
```

examples/hf_serialization.py
```diff
@@ -242,37 +325,36 @@ def hf_main(args):
...
+        del model
         gc.collect()
         device = utils.get_device()
+        model = AutoModelForCausalLM.from_pretrained(
+            args.input_directory,
+            config=model_config,
+            torch_dtype=dtype,
+            low_cpu_mem_usage=True,
+        )
+        model.to(device)
```
Eta0, line 343: `.to(device)` isn't in-place; this makes a copy on `device` and then discards it.
```suggestion
        model = AutoModelForCausalLM.from_pretrained(
            args.input_directory,
            config=model_config,
            torch_dtype=dtype,
            low_cpu_mem_usage=True,
        ).to(device)
```

examples/hf_serialization.py
```diff
@@ -304,6 +386,14 @@ def main():
...
         help="Validate serialization by running a test inference",
     )
+    parser.add_argument(
+        "--force",
+        action="store_true",
+        help=(
+            "Force upload serialized tensors to output_prefix even if they"
+            " already exist"
+        ),
+    )
```
Eta0, line 396: We should print a diagnostic message if serialization is skipped due to this flag being disabled so people can tell what is going on.

examples/hf_serialization.py
```diff
@@ -60,23 +110,40 @@ def serialize_model(
...
-    os.makedirs(model_directory, exist_ok=True)
-    dir_prefix = f"{model_directory}/{model_prefix}"
+    if model_prefix is None:
+        model_prefix = "model"
 
+    dir_prefix = f"{model_directory}/{model_prefix}"
+    config_file_exists, weights_file_exists = (
+        check_file_exists(f"{dir_prefix}-config.json"),
+        check_file_exists(f"{dir_prefix}.tensors"),
+    )
```
Eta0, line 124: There should be an error case if only one of these exists and `force` is not enabled, since selectively writing one of them will quite possibly break something. This might occur if they manually deleted the `.tensors` file and forgot to delete the `-config.json` file.

Eta0 (changes requested):

examples/hf_serialization.py
sangstar: The configuration file is still intended on being propagated. It's just now being converted to a `dict` and dumped as a JSON string. The only difference with this implementation is that it no longer tries to retrieve the `.to_json_file` method, as I was having trouble writing the file to S3 that way. In this case, it uses the `.to_dict` method instead and retains the case handling of checking directly if `config` is a `dict`.

Eta0: It's still saved, but it writes to a stream opened by `open_stream()` rather than using `.to_json_file()`, which presumably doesn't support object storage.

examples/hf_serialization.py
```diff
@@ -60,23 +110,40 @@ def serialize_model(
...
     if config is None:
         config = model
     if config is not None:
```
Eta0, line 127: This existed before the change, but is very fishy. `config` will never be `None` by the time it gets to the second `if` statement. Does this code actually work if `config` starts out as `None` (e.g. for `diffusers`)?

Eta0 (changes requested):

examples/hf_serialization.py
sangstar: Points taken. Will commit this.

examples/hf_serialization.py
sangstar: Okay, understood. Thanks for pointing out a forward-compatible change! Committing this.

examples/hf_serialization.py
sangstar: That's what I'd typically do, but I got a bit confused from the original handling in the source code for `load_model` that I thought reflected a general design preference:

```
def load_model(
    path_uri: str,
    model_class: Union[
        Type[PreTrainedModel], Type[ModelMixin], Type[ConfigMixin]
    ],
    config_class: Optional[
        Union[Type[PretrainedConfig], Type[ConfigMixin], Type[AutoConfig]]
    ] = None,
    model_prefix: Optional[str] = "model",
    device: torch.device = utils.get_device(),
    dtype: Optional[str] = None,
) -> torch.nn.Module:
    """
    Given a path prefix, load the model with a custom extension

    Args:
        path_uri: path to the model. Can be a local path or a URI
        model_class: The model class to load the tensors into.
        config_class: The config class to load the model config into. This must be
            set if you are loading a model from HuggingFace Transformers.
        model_prefix: The prefix to use to distinguish between multiple serialized
            models. The default is "model".
        device: The device onto which to load the model.
        dtype: The dtype to load the tensors into. If None, the dtype is inferred from
            the model.
    """

    if model_prefix is None:
        model_prefix = "model"
```

As I wasn't sure how this was meant to pass `None`, outside of explicitly putting `None` as its positional argument. I think I was just confused and I agree with your suggestion so happy to commit this.

examples/hf_serialization.py
sangstar: Good point. Will handle this.

examples/hf_serialization.py
sangstar: Not sure as this PR is only intended to titularly perform HF serialization.

examples/hf_serialization.py
sangstar: Agree, will add 👍

examples/hf_serialization.py
sangstar: Will add -- I assume `.rstrip("/")` does nothing if no `"/"` character exists on the string it's operating on, which is the desired functionality

examples/hf_serialization.py
Eta0: If the type on it is `Optional[str]`, that means that it's valid to put `None` as a positional argument. In this case, it makes more sense for it to be `str`, since the default can be clearly defined in the signature. Using `None` as a default makes sense if it's dynamic or based on other arguments in a way that can't be embedded in the call signature.

One reason why functions would be written like this otherwise is to not have to remember the default value if the calling code may or may not have a value to put there, and could conditionally need a default. In that case, the default would probably be better set further up the call chain, though.

examples/hf_serialization.py
sangstar: @harubaru mentioned this not being necessary in our discussion in the aforementioned thread, so it was the first thing I removed and hadn't revisited it since.

examples/hf_serialization.py
sangstar: Okay, I won't reinstantiate `model` in that case. Will resolve when this is fixed.

examples/hf_serialization.py
sangstar: Acknowledged.

examples/hf_serialization.py
Eta0: Yep, `.rstrip()` removes any-and-all trailing characters specified, so 0, 1, or more (e.g. `"dir"` or `"dir/"` or `"dir//"` → `"dir"`).

examples/hf_serialization.py
sangstar: Acknowledged. Will add this to batch.

examples/hf_serialization.py
sangstar: Will do so and resolve once fixed.

examples/hf_serialization.py
Eta0: Yes, however, this function is shared between both parts of this general `tensorizer` example for both the HF and DF codepaths. If it worked before due to the contents of the second `if` block, but doesn't work now, that would be a regression and need to be fixed to be compatible with both styles before it was merged, since we can't break it for one to improve the other.

examples/hf_serialization.py
sangstar: This should hopefully be resolved with the most recent commit.

examples/hf_serialization.py
harubaru: Tokenizer doesn't seem necessary for just serializing the weights to object storage unless we can upload the tokenizer onto object storage. If the HF ID is preserved then the tokenizer can just be fetched that way

harubaru (changes requested):
The Diffusers validation step should be the same as the validation for the language models. Though, if we want to keep the Diffusers support in this is another question as well since it seems like this PR's main objective is to serialize language models to object storage.

examples/hf_serialization.py
Eta0: This is resolved in #76, whenever that is merged

examples/hf_serialization.py
Eta0: Resolved in #76

sangstar:
> The Diffusers validation step should be the same as the validation for the language models. Though, if we want to keep the Diffusers support in this is another question as well since it seems like this PR's main objective is to serialize language models to object storage.

#76 from @Eta0 looks to have addressed this with `assert_model_equal`

examples/hf_serialization.py
sangstar: Tokenizers are now being saved in #76

Eta0 (approved):

PR #74, opened 2024-01-31
fix(examples): Miscellaneous `hf_serialization.py` fixes

Eta0:
# `hf_serialization.py` Fixes

Miscellaneous fixes and improvements for #73.

- Fixes calling `serialize_model` with a default `model_prefix`
- Fixes indentation for saved config files
- Fixes checking local files for emptiness
- Doesn't use `lazy_load=True` when it's not helpful
- Uses a more accurate timing function
- Refactors S3 stream credential propagation

### DIFF
--- a/examples/hf_serialization.py
+++ b/examples/hf_serialization.py
@@ -65,7 +65,7 @@ logger.addHandler(fh)
 def check_file_exists(file: str):
     """
     Check if file exists and is not empty. If the file is found locally,
-    it is checked for emptiness with `os.path.exists`. If the file is found
+    it is checked for emptiness with `os.stat`. If the file is found
     on S3, it is checked for emptiness by reading the first byte of the file.
 
     Args:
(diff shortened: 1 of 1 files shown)


examples/hf_serialization.py
```diff
@@ -120,19 +121,15 @@ def serialize_model(
...
             logger.info(f"Writing config to {config_path}")
-            with stream_io.open_stream(
-                config_path, "wb+", *s3_write_credentials
-            ) as f:
+            with _write_stream(config_path) as f:
                 config_dict = (
                     config.to_dict() if hasattr(config, "to_dict") else config
                 )
-                f.write(json.dumps(config_dict).encode("utf-8"))
+                f.write(json.dumps(config_dict, indent=2).encode("utf-8"))
```
sangstar, line 128: What's the purpose of this change?

sangstar (approved):
LGTM

examples/hf_serialization.py
Eta0: It was there before this PR. HuggingFace `config.json` files are typically indented.

PR #76, opened 2024-02-01
feat(examples): Support S3 for all `StableDiffusionPipeline` components

Eta0:
# Updated `diffusers` support in `hf_serialization.py`

More fixes and improvements for #73.

This change:
- Adds S3 upload capability for all `StableDiffusionPipeline` components
  - `text_encoder`, `vae` (updated), `unet` (updated), `scheduler` (new), `tokenizer` (new)
  - The `scheduler` and `tokenizer` are saved as `.zip` files containing the directory written by their `.save_pretrained()` methods
  - Requires nothing from HuggingFace Hub at deserialization time
- Adds S3 upload capability for `transformers` tokenizers
- Adjusts serialization to pass validation checks by using `include_non_persistent_buffers=False`
- Merges in the latest changes from `main` to support using `include_non_persistent_buffers=False` correctly
- Avoids re-initializing models from HF for no reason
- Cleans up misspelled/outdated CLI args and help
- Adds parameter weight validation for `diffusers` models
- Refactors `serialize_model` substantially
- Adds logging level command line arguments, and
- Shifts more outputs to use a logger

**(Outdated):** I left in the code to generate a test image through `diffusers` because it is good example code for this repository of how to re-assemble the components of an SD model and was good for testing that these changes work. It could be commented out or deleted later, or changed to be only enabled through a flag.
**(Update):** The test image generation code for diffusers is now commented out.


examples/hf_serialization.py
```diff
@@ -226,71 +303,86 @@ def df_main(args: argparse.Namespace) -> None:
...
+        )
 
-    pipeline.tokenizer.save_pretrained(output_prefix)
-    pipeline.scheduler.save_pretrained(output_prefix)
+    serialize_pretrained(
+        pipeline.tokenizer, output_prefix, "tokenizer", force=args.force
+    )
+    serialize_pretrained(
+        pipeline.scheduler, output_prefix, "scheduler", force=args.force
+    )
```
sangstar, line 324: So the purpose of `serialize_pretrained` is specifically to support saving artifacts like SD's `scheduler` and `tokenizer`? I assume this is cleaner as they're not configs nor modules.

examples/hf_serialization.py
```diff
@@ -305,15 +397,9 @@ def hf_main(args):
...
-        # I'm not absolutely certain whether or not TensorSerializer
-        # changes the model's state in any way
-        del model
-        gc.collect()
         device = utils.get_device()
-        model = AutoModelForCausalLM.from_pretrained(
-            args.input_directory, config=model_config, torch_dtype=dtype
-        ).to(device)
+        model = model.to(device)
+        gc.collect()
```
sangstar, line 402: Why is this necessary? I thought `del model` would be sufficient as I assumed Python abstracts away having to manually deal with garbage collection

sangstar (approved):
Dramatically improved! LGTM. One or two questions on your thought process

examples/hf_serialization.py
Eta0: Yes, pretty much. The scheduler can actually be saved as a single JSON file (around 300 to 400 bytes), so it could have its own special code to avoid a zip file and temporary directory, but it seemed unnecessary at this time.
In summary:
- `text_encoder` / `vae` / `unet` — tensors & `config.json`
  - We use `tensorizer` to save these
  - Potentially very, very large, so the optimizations in `tensorizer` are important
- `scheduler` — `config.json` only
  - Supports `.save_pretrained(dir)` and `.from_pretrained(dir)`
  - Roughly around 400 bytes, so downloader optimizations aren't a huge deal
- `tokenizer` — entire directory of files
  - Supports `.save_pretrained(dir)` and `.from_pretrained(dir)`
  - Roughly around 1.5 MB for SD 1.5, and compresses well, so zips work nicely
  - For models that use "[fast tokenizers](https://huggingface.co/learn/nlp-course/chapter6/3)," these can be saved as a single `tokenizer.json` file instead, but not all models support fast tokenizers

examples/hf_serialization.py
Eta0: `gc.collect()` is optional (and was there before this change), but the motivation is that this a decent time for the garbage collector to be invoked, since:
1. Memory was (potentially) just made available to free by moving the model out of CPU RAM to GPU RAM
2. Memory is about to be needed during deserialization
The call could be removed without any visible side-effects.

examples/hf_serialization.py
```diff
@@ -226,71 +303,86 @@ def df_main(args: argparse.Namespace) -> None:
...
+            scheduler=scheduler,
             safety_checker=None,
             feature_extractor=None,
             requires_safety_checker=False,
         ).to(device)
 
         prompt = "a photo of an astronaut riding a horse on mars"
         with torch.autocast(device.type):
-            pipeline(prompt).images[0].save("test.png")
+            deserialized_pipeline(prompt).images[0].save("test.png")
```
harubaru, line 377: I would get rid of the test inference code since the previous validation steps should suffice for verifying that the model weights are unchanged.

harubaru (changes requested):

examples/hf_serialization.py
Eta0: Commented it out in 032a12ff0b3e77ac724b719b38565096e0996580

Eta0: Also added validation for the deserialized tokenizer (and a small amount of validation for the deserialized scheduler) to compensate here, since this was the only part making use of those two components.

harubaru (approved):

PR #79, opened 2024-02-07
feat(serialization): Default to `plaid_mode=True` for CUDA tensors

Eta0:
# `plaid_mode=True` by Default

This change switches the default settings for `TensorDeserializer` to use `plaid_mode=True` when deserializing tensors to CUDA devices. `plaid_mode=False` remains the default for CPU devices, since `plaid_mode=True` is only available with CUDA.

`plaid_mode=True` has been confirmed to be equivalent in safety and functionality (other than performance) to `plaid_mode=False` since #51, so this is a safe, non-breaking change that should only improve performance for everyone using the default `TensorDeserializer` settings.

This change is also documented in the changelog, and includes an update to `tensorizer`'s version, putting it at **2.8.0**.

Additionally, all outdated and incorrect warnings about `plaid_mode` being unsafe have been snipped from the documentation, and explicit specifications of `plaid_mode=True` have been removed from all example code. Deserialization tests have also been adjusted accordingly.


tensorizer/serialization.py
```diff
@@ -1555,6 +1555,7 @@ def __init__(
                     for name, entry in self._metadata.items()
                     if filter_func(name)
                 }
+            num_tensors: int = len(self._metadata)
```
wbrown, line 1558: Looks like an undocumented change/fix?

wbrown (approved):
One comment, non-blocking.

tensorizer/serialization.py
Eta0: That's part of a small internal optimization that limits the number of `plaid_mode` buffers to the total number of tensors. I included it because there's no reason to allocate more than that anyway, and it would prevent potentially spiking RAM use up to 8x for code working on single-tensor files (for whatever reason) due to switching to `plaid_mode=True`.

PR #84, opened 2024-02-15
fix(examples): Ensure using non-cached endpoint when validating

sangstar:
This fix enforces `s3_endpoint = "object.ord1.coreweave.com"` for writes as well as reads in `hf_serialization.py` rather than `"accel-object.ord1.coreweave.com"` to prevent unwanted caching of model tensors. Caching tensors isn't helpful for a workflow with the sole purpose of serializing and validating model tensors, as this would just prevent overwritten tensors from being used for validation, instead using the older, cached ones. Additionally, updated `transformers` and `accelerate` dependencies to allow running `Mistral-7B`.

### DIFF
--- a/examples/hf_serialization.py
+++ b/examples/hf_serialization.py
@@ -4,6 +4,7 @@ import io
 import json
 import logging
 import os
+import re
 import tempfile
 import time
 import zipfile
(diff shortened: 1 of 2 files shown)


examples/hf_serialization.py
```diff
@@ -34,7 +34,7 @@
 
 s3_access_key_id = os.environ.get("S3_ACCESS_KEY_ID") or None
 s3_secret_access_key = os.environ.get("S3_SECRET_ACCESS_KEY") or None
-s3_endpoint = os.environ.get("S3_ENDPOINT_URL") or None
+s3_endpoint = "object.ord1.coreweave.com"  # To prevent caching old files
```
Eta0, line 37: This fails if you want to write to a bucket in `las1` or `lga1`.

Eta0 (changes requested):

Eta0 (approved):

PR #88, opened 2024-02-20
perf(serialization): Use optimal linear partitioning to divide segments

Eta0:
# Linear Partitioning

This improves the algorithm used to assign segments of a file to reader threads in #87.

An $O(n^{2} \cdot k)$ time, $O(n \cdot k)$ space (where $n$ = number of tensors, $k$ = requested number of readers) dynamic programming algorithm for optimal linear partitioning is used when $\frac{n \cdot (n + 1)}{2} \cdot k \lt 250000$, (less than 250000 comparison operations, about 250 tensors with 8 readers, takes about 140 ms).

When the number of comparison operations to compute the optimal partitioning scheme would be too large, it reverts to [an efficient greedy algorithm originally designed by Sean Werkema](https://www.werkema.com/2021/11/01/an-efficient-solution-to-linear-partitioning/) that takes $O(n)$ time and $O(k)$ space.

### DIFF
--- a/tensorizer/_linear_partition.py
+++ b/tensorizer/_linear_partition.py
@@ -124,9 +124,9 @@ def linear_partition(
                 best = best_prefix_partition[end, partitions_remaining]
                 if (
                     prefer_fewer
-                    and largest_partition <= best
-                    or not prefer_fewer
                     and largest_partition < best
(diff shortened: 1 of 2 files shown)


bchess (approved):
Great!

PR #94, opened 2024-02-27
fix(redis): relax Redis requirements to `>=4.5.5` from `==5.0.0`.

wbrown:
Customer has a hard requirement for `redis==4.5.5` due to bugs in `5.0.0` for their use case. This relaxes the requirement to `redis>=4.5.5` from `redis==5.0.0`.

Test cases pass for Redis with `4.5.5`.


Eta0 (approved):
From checking the [redis-py release notes](https://github.com/redis/redis-py/releases), we don't need features from `redis>4.5`, so this should be all good.

PR #97, opened 2024-02-27
fix(stream_io): Use region & signature version on S3 writes; use kwargs

Eta0:
# `s3_region_name` & `s3_signature_version` fixes

This change updates some of the code added in #93 to:
1. Use keyword arguments for `s3_region_name` and `s3_signature_version` in `stream_io.open_stream` instead of reordering the existing parameter list, which would be a breaking change, and
2. Use the `s3_region_name` and `s3_signature_version` parameters on object storage writes as well, as they were only applied to object storage reads.


wbrown (approved):

wbrown:
FYI, @lizzzcai -- some updates.

PR #98, opened 2024-02-28
perf(serialization) Reuse a pinned buffer when copying from gpu

bchess:
Similar to plaid mode, re-use one pinned buffer to handle the data transfer from GPU to CPU for serialization.

In main, serializing gpt-j-6B fp16 to nvme took 8.375s
In this branch, takes 4.796s

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3834,7 +3834,7 @@ class TensorSerializer:
         else:
             transferred = queue.Queue(maxsize=max_read_ahead)
 
-        biggest_tensor_bytes = max(t.nbytes for t in tensors)
+        biggest_tensor_bytes = max(t.element_size() * t.nelement() for t in tensors)
         staging_tensor = torch.empty(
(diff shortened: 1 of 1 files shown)


wbrown:
@bchess How does this affect tensors that are already on the CPU?

tensorizer/serialization.py
```diff
@@ -3834,6 +3834,11 @@ def _async_bulk_device_to_host_transfer(
         else:
             transferred = queue.Queue(maxsize=max_read_ahead)
 
+        biggest_tensor_bytes = max(t.nbytes for t in tensors)
```
Eta0, line 3837: `torch.Tensor.nbytes` is quite new (torch v2.1); this one will be more compatible, and we use it elsewhere in this file
```suggestion
        biggest_tensor_bytes = max(t.element_size() * t.nelement() for t in tensors)
```

Eta0 (changes requested):

Eta0 (approved):

PR #99, opened 2024-02-29
fix: Avoid regressions, add tests, better reference handling, misc. logic errors

Eta0:
# Various Fixes for #87

This PR contains various fixes and performance tweaks for #87.

Highlights:
- Open files and `CURLStreamFile`s can be used with `num_readers > 1`
- The linear partitioning algorithm has greatly improved performance and correctness
  - Around 25x faster in a common configuration
- `num_tensors=-1` following other partial reads works correctly
- Buffer pointers and artificial references to their sources are avoided in favour of `memoryview`s that directly hold strong references to their sources, lowering the likelihood of memory getting garbage collected while still in use
- Some cryptic error causing S3 reads with `num_readers > 1` to fail has been resolved
- Multiple separate reads share a thread pool
- HTTP `Range` end offsets are requested and handled correctly
- HTTP `Range`s have more thorough error checking in the `CURLStreamFile` constructor
- `CURLStreamFile`s try not to over-allocate file or pipe buffers when expecting to read a small range
- (Not directly related to #87) `stream_io.open_stream` is only allowed to open local files when passed a path-like object instead of a string
- `_bulk_load` permits requests for out-of-order keys
- `_bulk_load`'s reordering scheme has better asymptotic complexity (my original suggestion to use `heapq` was sub-optimal)
- PyTorch versions below 2.1 are supported again
- The original open file is used by the first reader thread to not waste its open connection and its read-ahead data
- `KeyboardInterrupt`s can trigger reader thread shutdown
- Reverted the extra error case in `verify_module`, since the conditions for it were removed in a later revision
- New tests have been added for CPU and S3 deserialization with `num_readers > 1`, also using the new reopening functionality for file-like objects

### DIFF
--- a/tensorizer/_linear_partition.py
+++ b/tensorizer/_linear_partition.py
@@ -1,7 +1,6 @@
-from typing import Iterable, List, Sequence
-
-import numpy as np
-
+import itertools
+import sys
+from typing import Iterable, List, Sequence, Tuple
 
 __all__ = ("partition",)
 
@@ -9,8 +8,7 @@ __all__ = ("partition",)
 def partition(
     weights: Sequence[int],
     partitions: int,
-    prefer_fewer: bool = True,
-    performance_threshold: int = 250000,
+    performance_threshold: int = 100,
 ) -> Iterable[slice]:
     """
     Partitions a sequence of weights into slices with balanced sums,
@@ -20,16 +18,12 @@ def partition(
         weights: Element weights to balance.
         partitions: The maximum number of slices to return.
             May return fewer if there are too few weights.
-        prefer_fewer: Enables returning fewer slices than requested
-            if the optimal partitioning scheme with at most `partitions`
-            partitions does not need exactly `partitions` slices.
-            For example, asking for 3 slices of ``[5, 2, 2]``
-            could return ``[[5], [2, 2]]``, since splitting again
-            to ``[[5], [2], [2]]`` doesn't improve the largest slice's sum.
-        performance_threshold: Limit for the number of comparisons that would
-            be used to calculate an optimal partitioning scheme. If more than
-            this many comparisons would be needed, an asymptotically faster
-            greedy approximation is used instead.
+        performance_threshold: Limit on the estimated time that would
+            be required to calculate an optimal partitioning scheme.
+            Not an exact measurement, but similar to milliseconds,
+            with an additional fuzzy bound on memory usage thrown in.
+            If this threshold is passed, an asymptotically faster,
+            low-memory greedy approximation is used instead.
 
     Returns:
         An iterable of ``slice`` objects denoting ranges of the original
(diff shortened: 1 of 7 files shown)


tensorizer/_wide_pipes.py
```diff
@@ -136,7 +138,7 @@ def _create_wide_pipe(pipe_attrs, size):
             )
 
         @contextlib.contextmanager
-        def widen_new_pipes():
+        def widen_new_pipes(max_size=None):
```
wbrown, line 141: Very nice optimization here.

tensorizer/_linear_partition.py
```diff
@@ -1,16 +1,14 @@
...
-import numpy as np
-
+import itertools
+import sys
+from typing import Iterable, List, Sequence, Tuple
 
 __all__ = ("partition",)
 
 
 def partition(
```
wbrown, line 8: This is pretty neat to read. :+1:

tensorizer/_linear_partition.py
```diff
@@ -51,108 +45,177 @@ def partition(
...
-                    # If prefer_fewer is true, this may result in fewer
-                    # partitions than originally requested, if adding more
-                    # wouldn't reduce the maximum partition size either way
-                    best_prefix_partition[end, partitions_remaining] = (
-                        largest_partition
+    inf: int = prefix_sums[-1] + 1
+    sentinel = (inf, inf)
+    memo = [sentinel] * ((n + 1) * partitions)
+    # Key function: end * partitions + preceding_parts
+    null = (0, 0)
```
wbrown, line 83: I know that this isn't a keyword in Python, but seeing `null` defined gives me hives. :joy:

tensorizer/_linear_partition.py
```diff
@@ -51,108 +45,177 @@ def partition(
...
+                elif dist == 1:
+                    start = right
+                else:
+                    start = left + (dist >> 1)
+
+        result = (best_start, best_weight)
+        memo[key] = result
+        return result
+
+    if partitions > 900:
```
wbrown, line 183: How'd we arrive at this number of partitions threshold?

tensorizer/stream_io.py
```diff
@@ -430,15 +461,15 @@ def _read_until(
             if ba is None:
                 rq_sz = goal_position - self._curr
                 if self._end is not None and self._curr + rq_sz > self._end:
-                    rq_sz = self._end - self._curr
+                    rq_sz = self._end - self._curr + 1
```
wbrown, line 464: Hmm, did we have an off by one here?

tensorizer/stream_io.py
```diff
@@ -531,7 +562,41 @@ def seek(self, position, whence=SEEK_SET):
...
+            A new `CURLStreamFile` identical to this one,
+            but with a different byte range.
+        """
+        clone = self.__class__(
+            uri=self._uri,
+            begin=begin,
+            end=end,
+            headers=self._headers,
+            buffer_size=self._buffer_size,
+        )
```
wbrown, line 597: Are we not passing `certificate_handling`?

wbrown (approved):
Some questions and comments, but great job. :+1:

tensorizer/_linear_partition.py
Eta0: Python's recursion depth limit (`sys.getrecursionlimit()`) is 1000 by default. The recursive algorithm used here has one layer of recursion per partition, so anything near the default recursion depth limit triggers raising the limit up to some still-pretty-safe upper limit (~6500), and anything beyond that uses special logic to avoid any deeper recursive calls that the algorithm would use. For tensorizer, these are very unlikely to ever come up, since the number of partitions is at most the number of readers (which is also at most the number of tensors), and having 1000 concurrent readers in one process is most likely a bad idea, and should revert to using the greedy algorithm for partitioning because of the performance threshold anyway.

I rewrote an earlier version of this to do manual stack management with Python lists instead of function calls to avoid recursion to never have to deal with this problem (and its performance was pretty good), but the logic in this one is a little bit more involved, so the actually-recursive one seems nicer to keep here unless we need to change it.

tensorizer/stream_io.py
Eta0: What ever do you mean? Of course we're passing `certificate_handling` 😇 (fixed in b6f2697555c86fe998d8ff9ff18b8e764f8f4612)

tensorizer/stream_io.py
Eta0: Indeed; I mentioned it in [a comment on #87](https://github.com/coreweave/tensorizer/pull/87#pullrequestreview-1908134052):

> HTTP `Range` requests have inclusive endpoints—e.g., a range of 0–1 is two bytes, not one. None of our code had been handling this correctly, but the multiple ways it got it wrong simultaneously had perfectly aligned to make it appear to behave correctly anyway. In #99, I changed it to handle end markers correctly, but I wonder if we should just formally change the `open_stream` and `CURLStreamFile` interfaces instead, so that their end parameters are exclusive of the final byte, like other slicing behaves in Python.

bchess (approved):
lg2m! Thanks :D

PR #101, opened 2024-03-01
perf(serialization) Make tensors contiguous via thread pool

bchess:
Making tensors contiguous() can be CPU-intensive, so throw that to a thread pool. Along the way, refactor the part of `_bulk_write()` that prepares tensors into a separate method `_prepare_tensors_for_write()`.

This handles currently three separate possible operations on each tensor:
* transferring it from cuda
* `clone().detach()` for when the tensor shares memory with another tensor and encryption is enabled
* making it `contiguous()` if it isn't
* .. plus a no-op

Each of these operations are done in a thread or threadpool. It's almost a list of futures, except `_async_bulk_device_to_host_transfer()` handles a list of tensors all at once. So instead I use iterators.

`examples/serialize.py` with gpt-j-6b (w/o CUDA) takes 17s on main, 8.8s after this change. I'm not sure how much of an impact this will have on actual use cases. But it can't hurt, and the refactor helps make this more maintainable imo.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3852,10 +3852,10 @@ class TensorSerializer:
         write_specs = collections.deque(write_specs)
         next_pos = self._file.tell()
         if _syscalls.has_fallocate() and self._fd:
-            size = sum(len(t.name) for t in tensors)
+            size = sum(len(w.name) for w in write_specs)
             size += sum(
-                t.tensor.element_size()
-                * t.tensor.nelement()
+                w.tensor.element_size()
+                * w.tensor.nelement()
                 * (not w.tensor.is_meta)
                 for w in write_specs
             )
@@ -3880,7 +3880,7 @@ class TensorSerializer:
                     w.tensor,
                     _synchronize=False,
                     _start_pos=next_pos,
-                    _temporary_buffer=w.temp_tensor,
+                    _temporary_buffer=w.temporary_buffer,
                 )
                 if w.callback is not None:
                     w.callback()
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3670,29 +3670,35 @@ class TensorSerializer:
         )
         return tensor_endpos
 
+    class _WriteSpec(typing.NamedTuple):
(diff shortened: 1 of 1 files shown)


tensorizer/serialization.py
```diff
@@ -3748,90 +3763,127 @@ def _interrupt_transfer():
...
-                interrupt_transfer,
-            ) = self._async_bulk_device_to_host_transfer(cuda_tensors)
-        else:
-            transferred = interrupt_transfer = None
-        del cuda_tensors
-
-        if self._encrypted:
-            shared = []
-            seen_addresses = set()
-            for t in reversed(tensors):
```
bchess, line 3791: @Eta0 I am not carrying forward this `reversed()`. Not sure if that was logic just to keep with the `popleft`-iness, or if there was some need to specifically clone() the earlier tensors instead of the later ones.

Eta0: The `reversed` is for correctness. Earlier tensors need to be cloned instead of later ones because of the following scenario where A and B share memory:

- A is handed to the writer pipeline with its original buffer
- A starts to be encrypted in place in the buffer
- B is cloned into a new buffer by the buffer-preparing pipeline and enqueued while this is happening
  - The state of B is now encrypted or partially-encrypted garbage data
- A finishes being written and decrypts in place
- B is handed to the writer
- B ends up being written double-encrypted

When the earlier tensors are cloned, the original buffer is only given to the writer thread (and opened up to being temporarily modified) *after* the clones have been made, so the clones are guaranteed to have the correct data.

Eta0: For what it's worth this situation also can't come up if all cloning happens before any writing, but ideally we are interleaving CPU-bound operations like cloning with the IO-bound operations like writing to maximize throughput on all resource types at once instead of batching them (which is the reason for the design of the device transfer thread)

wbrown:
@Eta0 @bchess Is this one in limbo?

bchess:
I'm ok to close this unmerged since it doesn't seem like there's any customer who would benefit from this.

PR #102, opened 2024-03-02
perf(stream_io): Don't use a temporary buffer for truncated reads

Eta0:
# Better Truncated Reads

This change avoids an unnecessary temporary buffer during truncated reads with `CURLStreamFile.readinto()`. It also avoids an off-by-one error that was sending reads for the last bytes of a stream down the truncated read code path (even if they were not truncated) if the `CURLStreamFile` had an `end` marker.


wbrown (approved):
Good one. :+1:

PR #104, opened 2024-03-02
perf(serialization): Preload CUDA

Eta0:
# Preload CUDA

This change adds a function to preload the lazily-initialized parts of `torch.cuda` in a background thread.
It calls a short CUDA function that does not hold the GIL (`torch.empty`) to allow some initialization to occur during tensorizer setup.

This PR is conservative about preloading and only does it exactly when `TensorDeserializer` is being requested to deserialize to a CUDA device, which doesn't allow for the full possible performance benefit of preloading with the current deserialization logic. It could potentially be started earlier, e.g. on module initialization, at the cost of unnecessarily initializing CUDA when only CPU serialization or deserialization is needed. The internal deserialization logic can alternatively be rearranged later to work better in the presence of a preloader (other optimizations like pipelining CUDA transfers as a separate thread may even improve this as a side effect).

## Avoiding `torch.Tensor.nbytes`

This change also removes a leftover reference to `torch.Tensor.nbytes` from #98, which was breaking compatibility with torch versions below 2.1. It also changes the serialization transfer thread to better avoid hanging the main thread if it errors.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3712,11 +3712,9 @@ class TensorSerializer:
         else:
             transferred = queue.Queue(maxsize=max_read_ahead)
 
-        biggest_tensor_bytes = max(
-            t.element_size() * t.nelement() for t in tensors
-        )
+        tensor_sizes = [t.element_size() * t.nelement() for t in tensors]
         staging_tensor = torch.empty(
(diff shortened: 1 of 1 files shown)


wbrown (approved):
Neat trick here. :+1:

PR #109, opened 2024-03-07
fix(serialization): Drop `atomics` dependency & make perf stats optional

Eta0:
# Locking instead of atomics, and optional `_perf_stats`

This change switches from using the `atomics` package for synchronized variable access to explicit locking, to avoid errors thrown by that package's deallocation code, and also for performance (amusingly). Additionally, the internal-only `_perf_stats` gathered during deserialization are now optional, and are only collected if the environment variable `TENSORIZER_ENABLE_PERF_STATS` is set.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -46,7 +46,6 @@ from typing import (
 import numpy
 import redis
 import torch
-from atomics.base import AtomicUint
 
 import tensorizer._crypt as _crypt
 import tensorizer._crypt_info as _crypt_info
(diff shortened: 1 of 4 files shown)


tensorizer/serialization.py
```diff
@@ -3984,11 +3999,12 @@ def write_state_dict(self, state_dict: Dict):
...
-    return dict(
-        cuda_to_device_secs=float(_perf_stats.cuda_to_device_millisecs.load())
-        / 1000.0,
-        cuda_bytes=_perf_stats.cuda_bytes.load(),
-        file_readinto_secs=float(_perf_stats.file_readinto_millisecs.load())
-        / 1000.0,
-        file_readinto_bytes=_perf_stats.file_readinto_bytes.load(),
-    )
+    if _perf_stats is None:
+        raise RuntimeError("Performance stats are not enabled")
```
wbrown, line 4003: Maybe mention how this can be enabled by setting `TENSORIZER_ENABLE_PERF_STATS"` in the `RuntimeError`.

wbrown (changes requested):
:+1:, one small comment to address.

tensorizer/serialization.py
Eta0: Since it is meant to be internal-only, I'm not sure we want to guide anyone on how to enable it unless they already know what they are doing with it (though on that same note, it should probably not be accessed in the `deserialize.py` example meant for public use, but rather something in the benchmarks folder). A source code comment next to the exception could be a compromise, though. What do you think?

bchess: I'm fine moving the current `deserialize.py` to `benchmarks/`, just as long as it stays intact :)

tensorizer/serialization.py
```diff
@@ -2333,16 +2334,20 @@ def _to_torch_parameter(
...
             tensor_on_device,
             requires_grad=gradient,
         )
-        _perf_stats.cuda_bytes.add(result.element_size() * result.nelement())
+        if _perf_stats and original_device != result.device:
+            duration = end - start
+            bytes_transferred = result.element_size() * result.nelement()
+            with _perf_stats.lock:
+                _perf_stats.tensor_to_device_ns += duration
+                _perf_stats.tensor_to_device_bytes += bytes_transferred
```
bchess, line 2350: At this point I'm pretty tempted to code the timer itself as a context within _perf_stats to keep the main code cleaner. Can be left as future work tho.

wbrown:
@Eta0 Doing an executive decision, and merging as `atomics` actually made this library stop working for a customer -- they do local development on Macs.

Going forward, we will be testing all changes on Macs in addition to Linux.

PR #110, opened 2024-03-07
perf: Add a syscall-based GIL-free address prefaulting function

Eta0:
# GIL-Free Memory Prefaulting

This change adds a `tensorizer_syscalls.prefault` function that can fault pages in memory for writing without holding the GIL, allowing prefaulting to be multithreaded. This accelerates `cudaHostRegister` performance, since it then does not need to fault pages itself.

The `prefault` function uses `madvise` if `MADV_POPULATE_WRITE` is available (Linux 5.14+), or `memset` otherwise, which incurs unnecessary writes, but is still fairly fast. Its arguments must be page-aligned for compatibility with the `madvise` syscall.

No usages of this function are added in the PR—only the function itself, to aid in development of better allocation schemes in subsequent PRs.

Example usage:

```py
import ctypes
import mmap
from concurrent.futures import ThreadPoolExecutor

from tensorizer._syscalls import prefault

num_threads: int = 16
buffer_size: int = 4 << 30

with ThreadPoolExecutor(max_workers=num_threads) as pool, mmap.mmap(
    -1, buffer_size, flags=mmap.MAP_PRIVATE
) as m:
    buf = (ctypes.c_ubyte * buffer_size).from_buffer(m)
    try:
        chunk_size: int = buffer_size // num_threads
        offsets = (
            ctypes.byref(buf, i) for i in range(0, buffer_size, chunk_size)
        )
        sizes = (chunk_size,) * num_threads
        for _ in pool.map(prefault, offsets, sizes):
            pass
    finally:
        del buf
```

### DIFF
--- a/tensorizer/_syscalls.py
+++ b/tensorizer/_syscalls.py
@@ -127,7 +127,7 @@ def _get_madvise():
 _madvise = _get_madvise()
 del _get_madvise
 
-_madv_populate_read: int = 23
+_madv_populate_write: int = 23
 
 
 def _can_prefault_with_madvise() -> bool:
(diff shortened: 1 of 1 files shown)


bchess (approved):
awesome!

PR #115, opened 2024-03-19
feat(serialization): Add serialization of structured objects

Eta0:
# Structured Object Serialization

This change adds serialization for structured objects in a way mirroring JSON.
`TensorSerializer.write_state_dict()` can now serialize nested dictionaries (or other mappings), nested lists (or other sequences), and arbitrary mixtures of dictionaries and lists. Nested structure is accessed at deserialization time either through repeated key lookup or the dedicated `TensorDeserializer.tree` method (open to other names).

## Usage

The usage is illustrated in the documentation for `TensorDeserializer.tree`:

```py
serializer = TensorSerializer(path)
nested_structure = {
    "model": {
        "layer": [
            {"weight": torch.tensor(...)},
            {"weight": torch.tensor(...)},
        ],
    },
    "lm_head": torch.tensor(...),
}
serializer.write_state_dict(nested_structure)
serializer.close()

deserializer = TensorDeserializer(path)

# All the layers can be accessed as nested mappings
assert list(deserializer.keys()) == ["model", "lm_head"]
assert list(deserializer["model"].keys()) == ["layer"]

# Note: when accessed without .tree(),
# the list turns into a mapping with int keys
assert list(deserializer["model"]["layer"].keys()) == [0, 1]

assert list(
    deserializer["model"]["layer"][0].keys()
) == ["weight"]

# The layers can be accessed with their
# original structure as well

from typing import Sequence, Mapping

tree = deserializer.tree()
assert isinstance(tree, Mapping)
assert isinstance(tree["model"]["layer"], Sequence)

# You can load the structure of a specific prefix
subtree = deserializer.tree(("model", "layer"))
assert isinstance(subtree, Sequence) and len(subtree) == 2
```

## Nested Structure Access

Accessing nested structures through direct key access always accesses them as (read-only) mappings, while the `TensorDeserializer.tree` method converts them to something matching their original type (`Sequence` or `Mapping`). The distinction is because it is possible to serialize a top-level sequence, but a `TensorDeserializer` is always a `Mapping` at its top level, and the `Mapping` and `Sequence` interfaces are not interchangeable. Additionally, mappings can have sparse integer keys (which aligns with the dynamics of a `filter_func`), while sequences must have sequential keys.

`TensorDeserializer` direct key access can only be done one path component at a time, and the `TensorDeserializer.keys()` method (and iterating over a `TensorDeserializer`) only returns top-level path components (i.e. the keys of the outermost structure). It does not walk the recursive structure. The behaviour of `TensorDeserializer.read_tensors()` is more like walking the structure, as it returns leaf tensors along with strings or tuples representing their full paths.

## Lazy, Eager, and Selective Loading

At the moment, the `TensorDeserializer.tree` method returns actual `dict` and `list` objects, and direct key access of a nested dictionary creates an actual `dict` and then makes it read-only by wrapping with a `types.MappingProxyType`. These are designed to be able to be changed to lazy view/proxy objects later, that can inherit `lazy_load` semantics from the `TensorDeserializer`. Subtrees are currently always (bulk-)loaded eagerly, and the `TensorDeserializer.tree` method exposes `prefix` and `filter_func` parameters to allow for more efficient selective loading.

As an extra interesting note, `TensorDeserializer.tree` is, as of this PR, the only function that allows non-stateful (unlike `.read_tensors()`) selective bulk loading *after initialization* for lazy-loaded deserializers, so it can also be used to effectively load multiple groups of keys from one lazy deserializer.

## Filter Functions

Filter functions have an extended signature when working with structured serialized tensors: `(tensor_path: str | Sequence[str | int]) -> bool`. Top-level `dict` string keys (previously the only keys possible for serialization and deserialization) are still given as single `str` objects, while any other structure (multiple components, or a single `int`) are always a `Sequence` of components. This allows writing filter functions like this:

```py
def filter_func(tensor_path) -> bool:
    # Load the key "thing", plus everything under the obj["array"][4]["subarray"] subtree
    if isinstance(tensor_path, str):
        return tensor_path == "thing"
    else:
        return tuple(tensor_path[:3]) == ("array", 4, "subarray")
```

Existing code that does not use structured serialization will only ever encounter plain `str` keys like before, so this is fully backwards-compatible.

Currently, `TensorDeserializer.load_into_module` does not support loading structured tensors. But since they can only be serialized by `TensorSerializer.write_state_dict`, which is incompatible with `TensorDeserializer.load_into_module` anyway, this is to be expected.

## Internal Format

Structured keys are stored as [`application/json-seq` (RFC 7464)](https://www.rfc-editor.org/rfc/rfc7464.html) arrays, which are single-line JSON arrays prefixed by a [`0x1E` ASCII record separator](https://www.compart.com/en/unicode/U+001E) character. There is no trailing newline required because arrays are self-delimited, as noted by RFC 7464. That is to say, the tensor accessed via `deserializer["a"]["b"]["c"][1][2][3]["x"]["y"]["z"]` would be encoded internally as a tensor named `\x1E["a","b","c",1,2,3,"x","y","z"]`.

The primary motivation for using `application/json-seq` as the format is that it provides both a straightforward indicator (in the form of a control character) that a field has special meaning and is to be parsed differently (as opposed to regular key names), has all the flexibility of JSON to represent arrays and mixed string and integer datatypes, is easily written and parsed with existing JSON libraries, is technically still an ASCII/UTF-8 string, so newer structured files can be opened with older deserializers (though without structure and the keys will look opaque), and it doesn't add significant size overhead, since most key data is already text, and wouldn't benefit from binary. Binary would still be preferable for parsing most integers, so future major revisions of the tensorizer format could implement a better custom list format while doing away with plain string key compatibility.

This is technically a breaking change for any tensor names that previously started with the control code `0x1E`, which seems acceptable.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -12,7 +12,6 @@ import functools
 import hashlib
 import io
 import itertools
-import json
 import logging
 import operator
 import os
@@ -53,6 +52,7 @@ import tensorizer._crypt as _crypt
 import tensorizer._crypt_info as _crypt_info
 import tensorizer._linear_partition as _linear_partition
 import tensorizer._syscalls as _syscalls
+import tensorizer._tensor_path as _tensor_path
 import tensorizer.stream_io as stream_io
 import tensorizer.utils as utils
 from tensorizer._crypt._cgroup_cpu_count import (
@@ -61,6 +61,11 @@ from tensorizer._crypt._cgroup_cpu_count import (
 from tensorizer._internal_utils import Chunked as _Chunked
 from tensorizer._internal_utils import _variable_read
 from tensorizer._NumpyTensor import _NumpyTensor
+from tensorizer._tensor_path import (
+    _TensorPath,
+    _TensorPathComponent,
+    _TensorPathRegistry,
+)
 
 __all__ = [
     "TensorSerializer",
@@ -104,10 +109,6 @@ cpu_count: int = _effective_cpu_count()
 class _SupportsBool(typing.Protocol):
     def __bool__(self) -> bool: ...
 
-
-# Tensor paths are made up of strings (for mapping keys)
-# and integers (for array indices)
-_TensorPathComponent: "typing.TypeAlias" = Union[str, int]
 
 # Filter functions take either a single string (for a simple top-level dict key)
 # or a tensor path (a sequence of path components, each a string or integer).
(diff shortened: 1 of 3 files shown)


tensorizer/serialization.py
```diff
@@ -170,6 +189,160 @@ class TensorHash:
...
 
+class _TensorPath(tuple):
+    def serialized_(self) -> bytes:
+        if self.is_str_:
+            return self[0].encode("utf-8")
+        else:
+            # application/json-seq format
+            return b"\x1e" + json.dumps(
+                self, indent=None, ensure_ascii=True, separators=(",", ":")
+            ).encode("ascii")
```
wbrown, line 200: Why are we `ascii` encoding the JSON here?

wbrown (commented):
One small comment re: `ascii` encoding.

One thing we may want to start considering is breaking up `serialization.py`. It's a pretty huge file now.

tensorizer/serialization.py
Eta0: It needs to be converted from `str` to `bytes`, so it needs to be encoded to something, and as JSON here it's already ASCII (all non-ASCII Unicode characters in json-seq strings are backslash-escaped), so it's the most correct option.

Eta0:
> One small comment re: `ascii` encoding.
> 
> One thing we may want to start considering is breaking up `serialization.py`. It's a pretty huge file now.

@wbrown I've refactored it a bit in dc7da64209516a00788f296d95fa3e9090107e7f to move a lot of the new logic to its own file (`_tensor_path.py`).

PR #122, opened 2024-04-11
feat(serialization): Use a dynamic `num_readers` by default

Eta0:
# Multi-reader deserialization by default

This switches the default number of concurrent readers during deserialization from 1 to a dynamic value based on the file being  deserialized, and the system resources.

The logic to choose how many readers to use by default is:
- Anything that the code doesn't know how to reopen uses 1, otherwise,
- Local files use 8, otherwise,
- Anything using hash verification uses 4, otherwise,
- Anything else uses 2
- If a `CURLStreamFile`'s headers do not include `Accept-Ranges: bytes`, use 1
- Use fewer readers if the number picked is expected to overflow RAM
  - E.g. if only the top 4 largest tensors are expected to fit in memory at one time, don't default to more than 4 readers
  - This check is still bypassed if a specific `num_readers` is used
- Never use more readers than there are tensors in the file

An extra check is also added that reopened files have a matching [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) to their original file, though it only emits a log statement if it fails (since ETags are not strictly guaranteed to be stable). This is because the initial reader could hypothetically fetch an older cached version of a resource when computing all the metadata, while subsequent readers get a newer version where their expected offsets are no longer valid. Since that would be a very confusing error if it comes up, a log statement could help figure it out.

This also pre-emptively raises an error if `num_readers` is specifically requested to be greater than 1 and the server does not give the `Accept-Ranges: bytes` header, where if `num_readers` is dynamic, it simply resets to 1.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -1692,31 +1692,9 @@ class TensorDeserializer(
                 )
 
             if dynamic_num_readers:
-                if isinstance(
-                    getattr(self._file, "raw", self._file), io.FileIO
-                ):
-                    num_readers = 8
-                elif self._verify_hash:
-                    num_readers = 4
-                else:
-                    num_readers = 2
-                if is_cuda:
-                    free_ram = psutil.virtual_memory().available
-                    allowed_ram = free_ram - (10 << 20)
-                    tensor_sizes = sorted(
-                        (
-                            t.deserialized_length
-                            for t in self._metadata.values()
-                        ),
-                        reverse=True,
-                    )
-                    num_readers = min(num_readers, len(tensor_sizes)) or 1
-                    while (
-                        num_readers > 1
-                        and sum(tensor_sizes[:num_readers]) > allowed_ram
-                    ):
-                        num_readers -= 1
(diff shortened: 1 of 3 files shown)


tensorizer/serialization.py
```diff
@@ -1681,30 +1684,86 @@ def __init__(
...
+                        ),
+                        reverse=True,
+                    )
+                    num_readers = min(num_readers, len(tensor_sizes)) or 1
+                    while (
+                        num_readers > 1
+                        and sum(tensor_sizes[:num_readers]) > allowed_ram
+                    ):
+                        num_readers -= 1
+                    del tensor_sizes
```
bchess, line 1719: nit; pull out into a separate method

bchess (approved):
Great!

tensorizer/serialization.py
Eta0: Done in 452e99e842cc46261a787fcae7d2c9360320d1ad

wbrown (approved):
Excellent!

PR #123, opened 2024-04-11
fix(stream_io): Use exclusive end for `CURLStreamFile`s + misc bugs

Eta0:
# Exclusive-end `CURLStreamFile` ranges

This change switches the `end` parameter for `CURLStreamFile`s and `open_stream` to mean one-past-the-end, since that matches Python slicing, and is what most Python programmers would expect.

Evidence for this is that our implementation of `CURLStreamFile.read()` and multi-reader deserialization were incorrect because even we all assumed the `end` parameter meant one-past-the-end while developing tensorizer until a bug arose with it during v2.9.0 development. The fix then was to make it match HTTP range behaviour of `end` being inclusive, but that is enough of an unintuitive footgun that it will very likely lead to more bugs later on.

It was also arguably a breaking change, as code already using `CURLStreamFile`s' `end` parameter had their behaviour change.

This PR also fixes bugs with `CURLStreamFile.__del__` on errors, and offset tracking when using `CURLStreamFile.read()` with no argument. It also adds more test cases verifying that invalid ranges raise errors correctly in the `CURLStreamFile` constructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `end` applies only to HTTP(S) & S3 streams, for which it is interpreted
       as the `start` and `end` parameters for the created `CURLStreamFile`
       object
-      - `end` is inclusive, like in an HTTP byte range request,
-        not one-past-the-end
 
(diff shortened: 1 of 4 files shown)


bchess (approved):
Yes, thank you for this

wbrown (approved):
Good change, even if I think Python's weird. ;)

PR #124, opened 2024-04-17
perf(serialization): Calculate dynamic `num_readers` counts efficiently

Eta0:
# Better dynamic `num_readers` calculation

This is a small change to make `TensorDeserializer._choose_dynamic_num_readers` more efficient for large numbers of tensors by using `heapq.nlargest` instead of `sorted`, prefix sums instead of repeated summation, and `bisect` instead of linear search.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3,6 +3,7 @@
 # Fast torch module/model serialization/deserialization     (c) 2023 Coreweave
 ##############################################################################
 import abc
+import bisect
 import collections.abc
 import concurrent.futures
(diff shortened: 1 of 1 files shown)


wbrown:
@Eta0 Was this actually a problem? Was it slow enough to affect things?

Eta0:
Not particularly, which is why I did this as a quick thing after the v2.9.0 release instead of including it for that. It mostly only matters if you use an extremely large number of (probably small) tensors in a single file, because of the worse time complexity for sorting metadata.

PR #127, opened 2024-04-18
[3.0-dev] Major refactor of serialization in prep for 3.0

bchess:
(note this PR is just for merging into the `v3.0.0-dev` branch, not main)

# file schema
The file schema has changed. Immediately following the metadata block is the full block of all header structures.  

After the metadata and headers is the block of raw tensor data. The block of tensor data begins on a 4k page-aligned boundary (relative to start of file). Each individual tensor begins on an 8-byte-aligned boundary.

Basically:
```
file_header
metadata_size
metadata0
...
metadataN
header0
...
headerN
tensor0 (file_offset % 4096 == 0)
tensor1 (file_offset % 8 == 0)
...
tensorN (file_offset % 8 == 0)
```
We may want to combine the metadata and the header entries, but they are left separate for now for easier backwards compatibility.

## hash computation
Previously, header hash computations included the hash fields themselves, presumed to be zeroed. This has been changed so the hash computation segments omit the hash fields.  ~This breaks `verify_hashes=True` with 2.x files for now. We should prob fix prior to release~ (This is now fixed, we can deser 2.x files with verify_hashes)

# bulk_write refactor
`_bulk_write()` has been broken up into a series of smaller functions that each perform some operation on the tensors. `_WriteSpec` carries the state of each tensor as it gets processed. Many ops are threaded and tracked as Futures (`_WriteSpec.tensor_data_task`). Dependent steps do Future chaining.

## tensor_write()
`tensor_write()`, e.g. "incremental mode", is still supported. (It is implemented as a call to `_bulk_write([tensor])`). There are new tests to validate its functionality.

### max_tensors
Because the new file schema must know ahead of time where to start writing tensor data, there must be a certain amount of pre-allocated space for the metadata and headers. Thus you must specify `max_tensors` to `TensorSerializer.__init__` If you plan on callling `tensor_write()` more than once. This does a wild-ass-guess estimate on a good amount of header space to reserve. In some cases, like really long tensor names, it could still be an underestimate. You will get a RuntimeError() if you call `write_tensor()` and there isn't enough pre-allocated space.

You do not need to set `max_tensors` if you just call `tensor_write()` once.

# Performance
Note that while these changes are mainly motivated to improve performance, I did not spend much time in this PR to measure and optimize performance. There may be regressions, but we should be able to resolve them (and beyond!) before release.

### DIFF
--- a/tensorizer/_futuregroup.py
+++ b/tensorizer/_futuregroup.py
@@ -51,7 +51,13 @@ def _future_wait_and_raise(
     # Wait on a list of futures with a timeout. Raise any exceptions, including TimeoutErrors.
     # otherwise return the list of results in the same order as the input futures.
     results = []
-    fs = concurrent.futures.wait(futures, timeout=timeout)
+    flattened_futures = []
+    for f in futures:
+        if isinstance(f, _FutureGroup):
+            flattened_futures.extend(f.futures)
+        else:
+            flattened_futures.append(f)
+    fs = concurrent.futures.wait(flattened_futures, timeout=timeout)
     for f in fs.done:
         # if the future has an exception, this will raise it
         results.append(f.result())
--- a/tensorizer/_futuregroup.py
+++ b/tensorizer/_futuregroup.py
@@ -3,7 +3,7 @@ from typing import Any, List, Optional, Sequence, Union
 
 
 class _FutureGroup:
-    def __init__(self, futures: Sequence[concurrent.futures.Future]):
+    def __init__(self, futures):  # type: (Sequence[_Future]) -> None
         self.futures = futures
 
     def cancel(self) -> bool:
--- a/tensorizer/_futuregroup.py
+++ b/tensorizer/_futuregroup.py
@@ -1,0 +1,61 @@
+import concurrent.futures
+from collections.abc import Callable
+from typing import Any, Optional, Sequence
+
+
+class _FutureGroup(concurrent.futures.Future):
+    def __init__(self, futures: Sequence[concurrent.futures.Future]):
+        self.futures = futures
+
+    def cancel(self) -> bool:
+        result = True
+        for f in self.futures:
+            result = result and f.cancel()
+        return result
+
+    def cancelled(self) -> bool:
+        return all(f.cancelled() for f in self.futures)
+
+    def running(self) -> bool:
+        return any(f.running() for f in self.futures)
+
+    def done(self) -> bool:
+        return all(f.done() for f in self.futures)
+
+    def result(self, timeout=None) -> Sequence[Any]:
+        return _future_wait_and_raise(self.futures, timeout=timeout)
+
+    def exception(self, timeout=None) -> Optional[BaseException]:
+        for f in self.futures:
+            exc = f.exception(timeout=timeout)
+            if exc:
+                return exc
+        return None
+
+    def add_done_callback(self, _):
+        raise NotImplementedError()
+
+    def set_running_or_notify_cancel(self) -> bool:
+        raise NotImplementedError()
+
+    def set_result(self, _) -> None:
+        raise NotImplementedError()
+
+    def set_exception(self, _) -> None:
+        raise NotImplementedError()
+
+
+def _future_wait_and_raise(
+    futures: Sequence[concurrent.futures.Future], timeout=None
+) -> Sequence[Any]:
+    # Wait on a list of futures with a timeout. Raise any exceptions, including TimeoutErrors.
+    # otherwise return the list of results in the same order as the input futures.
+    results = []
+    fs = concurrent.futures.wait(futures, timeout=timeout)
+    for f in fs.done:
+        # if the future has an exception, this will raise it
+        results.append(f.result())
+    for f in fs.not_done:
+        # force raise of TimeoutError
+        results.append(f.result(0))
+    return results
--- a/tensorizer/_futuregroup.py
+++ b/tensorizer/_futuregroup.py
@@ -1,9 +1,8 @@
 import concurrent.futures
-from collections.abc import Callable
-from typing import Any, Optional, Sequence
+from typing import Any, List, Optional, Sequence, Union
 
 
-class _FutureGroup(concurrent.futures.Future):
+class _FutureGroup:
     def __init__(self, futures: Sequence[concurrent.futures.Future]):
         self.futures = futures
 
@@ -22,8 +21,8 @@ class _FutureGroup(concurrent.futures.Future):
     def done(self) -> bool:
         return all(f.done() for f in self.futures)
 
-    def result(self, timeout=None) -> Sequence[Any]:
-        return _future_wait_and_raise(self.futures, timeout=timeout)
+    def result(self, timeout=None) -> None:
+        _future_wait_and_raise(self.futures, timeout=timeout)
 
     def exception(self, timeout=None) -> Optional[BaseException]:
         for f in self.futures:
@@ -32,36 +31,26 @@ class _FutureGroup(concurrent.futures.Future):
                 return exc
         return None
 
-    def add_done_callback(self, _):
-        raise NotImplementedError()
 
-    def set_running_or_notify_cancel(self) -> bool:
-        raise NotImplementedError()
-
-    def set_result(self, _) -> None:
-        raise NotImplementedError()
-
-    def set_exception(self, _) -> None:
-        raise NotImplementedError()
+_Future = Union[_FutureGroup, concurrent.futures.Future]
 
 
-def _future_wait_and_raise(
-    futures: Sequence[concurrent.futures.Future], timeout=None
-) -> Sequence[Any]:
+def _future_wait_and_raise(futures: Sequence[_Future], timeout=None) -> None:
     # Wait on a list of futures with a timeout. Raise any exceptions, including TimeoutErrors.
-    # otherwise return the list of results in the same order as the input futures.
-    results = []
-    flattened_futures = []
-    for f in futures:
+
+    flattened_futures: List[concurrent.futures.Future] = []
+    futures = list(futures)
+    while futures:
+        f = futures.pop()
         if isinstance(f, _FutureGroup):
-            flattened_futures.extend(f.futures)
+            futures.extend(f.futures)
         else:
             flattened_futures.append(f)
+
     fs = concurrent.futures.wait(flattened_futures, timeout=timeout)
     for f in fs.done:
         # if the future has an exception, this will raise it
-        results.append(f.result())
+        f.result()
     for f in fs.not_done:
         # force raise of TimeoutError
-        results.append(f.result(0))
-    return results
+        f.result(0)
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -4201,7 +4201,7 @@ class TensorSerializer:
             w.set_min_file_version_number(OPAQUE_TENSORIZER_VERSION)
 
     @staticmethod
-    def _do_clone(write_spec, dependency: Optional[concurrent.futures.Future]):
+    def _do_clone(write_spec, dependency: Optional[_Future]):
         if dependency is not None:
             dependency.result(_TIMEOUT)
         write_spec.tensor = write_spec.tensor.clone().detach()
@@ -4234,7 +4234,7 @@ class TensorSerializer:
                 ]
             )
 
-            clone_tasks = []
+            clone_tasks: List[_Future] = []
             for w in shared_write_specs[1:]:
                 clone_tasks.append(
                     self._computation_pool.submit(
@@ -4401,7 +4401,7 @@ class TensorSerializer:
         for w in write_specs:
             old_tensor_data_task = w.tensor_data_task
 
-            hash_tasks = []
+            hash_tasks: List[_Future] = []
             if w.include_crc32:
                 crc32_task = self._computation_pool.submit(
                     compute_crc32, w, old_tensor_data_task
@@ -4418,7 +4418,7 @@ class TensorSerializer:
                 self._jobs.extend(hash_tasks)
 
     def _do_encryption(self, write_specs: Sequence[_WriteSpec]) -> None:
-        def encrypt(write_spec, dependency: _Future):
+        def encrypt(write_spec, dependency: Optional[_Future]):
             if dependency is not None:
                 dependency.result(_TIMEOUT)
             try:
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -8,6 +8,7 @@ import collections
 import collections.abc
 import concurrent.futures
 import contextlib
+import ctypes
 import dataclasses
 import enum
 import functools
@@ -3778,7 +3779,7 @@ class TensorSerializer:
             self.module_index = module_index
             self.tensor_type: TensorType = tensor_type
             self.name = _TensorPath.wrap_(name)
-            self.dtype: Optional[str] = None  # _prepare_for_write_numpy_tensor
+            self.dtype: Optional[str] = None  # _prepare_for_write_dtype
             self.shape = tensor.size()
             self.data_length = tensor.nbytes
             # self.file_offset  # intentionally omitted, handled by _write_headers()
@@ -3789,9 +3790,6 @@ class TensorSerializer:
             )
 
             # Additional payloads that get set and used during the prepare_for_write procedures
-            self.numpy_tensor: Optional[_NumpyTensor] = (
-                None  # $et in _prepare_for_write_numpy_tensor
-            )
             self.header: Optional[_TensorHeaderSerializer] = (
                 None  # $et in _prepare_for_write_headers
             )
@@ -3804,6 +3802,13 @@ class TensorSerializer:
             # e.g. cuda transfer, make_contiguous, hashing, encryption, writing, or decryption.
             # They are often chained from one step of the process to the next
             self.tensor_data_task: Optional[_Future] = None
+
+        @property
+        def tensor_memoryview(self) -> memoryview:
+            nbytes = self.tensor.element_size() * self.tensor.nelement()
+            return memoryview(
+                (ctypes.c_char * nbytes).from_address(self.tensor.data_ptr())
+            )
 
         def set_min_file_version_number(self, version_number):
             self.min_file_version = max(self.min_file_version, version_number)
@@ -3841,8 +3846,7 @@ class TensorSerializer:
         try:
             self._prepare_for_write_contiguous(write_specs)
             self._prepare_for_write_meta(write_specs)
-            self._prepare_for_write_numpy_tensor(write_specs)
-            self._prepare_for_write_opaque(write_specs)
+            self._prepare_for_write_dtype(write_specs)
             if self._encrypted:
                 self._prepare_for_write_encryption(write_specs)
             self._prepare_for_write_headers(write_specs)
@@ -4172,33 +4176,31 @@ class TensorSerializer:
                 make_contiguous, w, w.tensor_data_task
             )
 
-    def _prepare_for_write_numpy_tensor(
-        self, write_specs: Sequence[_WriteSpec]
-    ):
-        for w in write_specs:
-            # all futures are resolved here. This step is not multi-threaded.
-            if w.tensor_data_task is not None:
-                w.tensor_data_task.result(_TIMEOUT)
-                w.tensor_data_task = None
-            w.numpy_tensor = _NumpyTensor.from_tensor(w.tensor)
-            w.dtype = w.numpy_tensor.numpy_dtype
-            if w.numpy_tensor.data.data.nbytes != w.tensor.nbytes:
-                raise ValueError(
-                    f"Cannot serialize tensor {w.name!r}:"
-                    f" buffer size of underlying memory ({w.numpy_tensor.data.data.nbytes})"
-                    f" doesn't match reported size ({w.tensor.nbytes})"
-                )
+    def _prepare_for_write_dtype(self, write_specs: Sequence[_WriteSpec]):
+        torch_dtype_to_numpy_dtype_cache: Dict[str, str] = {}
 
-    def _prepare_for_write_opaque(
-        self, write_specs: Sequence[_WriteSpec]
-    ) -> None:
         for w in write_specs:
-            if not w.numpy_tensor.is_opaque:  # type: ignore
-                continue
-            # The datatype name needs to contain both the numpy dtype that the
-            # data is serialized as and the original torch dtype.
-            w.dtype += OPAQUE_DTYPE_SEP + w.numpy_tensor.torch_dtype  # type: ignore
-            w.set_min_file_version_number(OPAQUE_TENSORIZER_VERSION)
+            tensor_dtype_str = str(w.tensor.dtype)
+            if _NumpyTensor._is_asymmetric(w.tensor.dtype):
+                # is opaque
+                w.dtype = (
+                    f"<V{w.tensor.element_size():d}"
+                    + OPAQUE_DTYPE_SEP
+                    + tensor_dtype_str
+                )
+                w.set_min_file_version_number(OPAQUE_TENSORIZER_VERSION)
+            else:
+                w.dtype = torch_dtype_to_numpy_dtype_cache.get(
+                    tensor_dtype_str, None
+                )
+                if w.dtype is None:
+                    torch_dtype_to_numpy_dtype_cache[tensor_dtype_str] = (
+                        w.dtype
+                    ) = (
+                        torch.tensor((), dtype=w.tensor.dtype, device="cpu")
+                        .numpy()
+                        .dtype.str
+                    )
 
     @staticmethod
     def _do_clone(write_spec, dependency: Optional[_Future]):
@@ -4250,7 +4252,6 @@ class TensorSerializer:
                 w.tensor_data_task = _FutureGroup(clone_tasks)
 
         for w in write_specs:
-            assert w.numpy_tensor is not None
             w.include_crc32 = False
 
             if w.data_length == 0:
@@ -4263,7 +4264,7 @@ class TensorSerializer:
                 w.tensor_data_task.result(_TIMEOUT)
                 w.tensor_data_task = None
 
-            tensor_memory: memoryview = w.numpy_tensor.tensor_memory
+            tensor_memory: memoryview = w.tensor_memoryview
             chunked = _Chunked(
                 total_size=tensor_memory.nbytes,
                 chunk_size=self._crypt_chunk_size,
@@ -4383,9 +4384,7 @@ class TensorSerializer:
             if dependency is not None:
                 dependency.result(_TIMEOUT)
             header_crc32 = write_spec.header.compute_crc32()
-            crc32 = zlib.crc32(
-                write_spec.numpy_tensor.tensor_memory, header_crc32
-            )
+            crc32 = zlib.crc32(write_spec.tensor_memoryview, header_crc32)
             write_spec.header.add_crc32(crc32)
 
         def compute_sha256(
@@ -4395,7 +4394,7 @@ class TensorSerializer:
             if dependency is not None:
                 dependency.result(_TIMEOUT)
             sha256 = write_spec.header.compute_sha256()
-            sha256.update(write_spec.numpy_tensor.tensor_memory)
+            sha256.update(write_spec.tensor_memoryview)
             write_spec.header.add_sha256(sha256.digest())
 
         for w in write_specs:
@@ -4480,7 +4479,7 @@ class TensorSerializer:
                 bytes_written = 0
             else:
                 bytes_written = self._pwrite(
-                    write_spec.numpy_tensor.tensor_memory,
+                    write_spec.tensor_memoryview,
                     write_spec.header.data_offset,  # type: ignore
                     verify=write_spec.header.data_length,  # type: ignore
                 )
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -4201,6 +4201,7 @@ class TensorSerializer:
         if dependency is not None:
             dependency.result(_TIMEOUT)
         write_spec.tensor = write_spec.tensor.clone().detach()
+        write_spec.numpy_tensor = _NumpyTensor.from_tensor(write_spec.tensor)
 
     def _prepare_for_write_encryption(
         self, write_specs: Sequence[_WriteSpec]
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3755,7 +3755,7 @@ class TensorSerializer:
         def __init__(
             self,
             module_index: int,
-            name: str | _TensorPath,
+            name: Union[str, _TensorPath],
             tensor_type: TensorType,
             tensor: torch.Tensor,
         ):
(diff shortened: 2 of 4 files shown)


tensorizer/serialization.py
```diff
@@ -3700,529 +3766,152 @@ def write_tensor(
...
+            self.include_crc32 = True
+            self.include_sha256 = True
+            self.crypt_info: Optional[_crypt_info.CryptInfo] = (
+                None  # _prepare_for_write_encryption
             )
 
-        tensor: numpy.ndarray = numpy_tensor.data
-        tensor_memory: memoryview = numpy_tensor.data.data
-        tensor_size: int = tensor.nbytes
-        if tensor_memory.nbytes != tensor_size:
```
harubaru, line 3785: Where did this assertion go?

tensorizer/serialization.py
```diff
@@ -3700,529 +3766,152 @@ def write_tensor(
...
-        self._synchronize_pools()
-        self._file.seek(next_pos)
-        self._sync_prologue_state()
+        if cuda_executor is not None:
+            cuda_executor.shutdown(wait=True)
 
     def write_module(
         self,
         m: torch.nn.Module,
-        remove_tensors: bool = False,
```
harubaru, line 4225: What are the consequences of removing this argument? Do we know if anyone is using this?

tensorizer/serialization.py
```diff
@@ -2858,19 +2919,24 @@ def _copy_thread(
...
-                    raise ValueError("Unexpected empty header")
+                if unsafe_self._headers is None:
+                    header = _TensorHeaderDeserializer.from_io(
+                        unsafe_self._file_header.version_number,
+                        file_,
+                        zero_hashes=True,
+                        check_crypt_info=unsafe_self._has_crypt_info,
+                    )
+                    if header is None:
+                        raise KeyError("Unexpected empty header")
```
harubaru, line 2930: Why is this a `KeyError` instead of a `ValueError`?

tensorizer/serialization.py
```diff
@@ -3266,6 +3333,7 @@ def __init__(
         *,
         encryption: Optional[EncryptionParams] = None,
         limit_cpu_concurrency: Optional[int] = None,
+        max_tensors: Optional[int] = None,
```
harubaru, line 3336: Can there be added docs on this?

harubaru:
## Feedback

The code looks pretty good and it also works pretty good too. I have tested this against various serialization tasks by serializing models for inference which did not result in a failure. The comments that I have left on this are mainly to get a better understanding behind the intent of some of the changes in this PR even if it was as little as changing an error return value.

That being said, the changes to the file schema that this PR provides are nice such as aligning tensor data on fixed boundaries or the decomposition of `bulk_write()`. Though, I am a bit curious about the addition of `max_tensors`. Is `max_tensors` a necessary argument to have because we support `write_tensor()`?

Additionally, it would be helpful to understand `max_tensors` more with a code sample that demonstrates its necessity, just like the code samples in the other docstrings.

## Testing

A good portion of the changes in this PR and the PR's intent is changing the file schema. I have spent a good portion of my time reviewing this PR by testing the changes that were implement. The PR works with deserializing older 2.x artifacts and I have gathered benchmarking data from serializing and deserializing GPT-J 6B for 50 trials which did not show a significant performance difference between this PR and 2.9.0 (or at least: I don't think it is a significant difference.)

image

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+                        self._do_clone, w, clone_dependencies
+                    )
+                )
+                w.user_owns_tensor_data = False
+
+            shared_write_specs[0].tensor_data_task = _FutureGroup(
+                clone_tasks + clone_dependencies.futures
+            )
+            for w in shared_write_specs[1:]:
+                w.tensor_data_task = _FutureGroup(clone_tasks)
```
Eta0, line 4285: I would choose the non-cloned tensor to be the last one, because its `tensor_data_task` will block the longest, so that would unblock some processing for a subsequent stage sooner, considering the access pattern of `write_specs` is consistently first-to-last.
```suggestion
            for w in shared_write_specs[:-1]:
                clone_tasks.append(
                    self._computation_pool.submit(
                        self._do_clone, w, clone_dependencies
                    )
                )
                w.user_owns_tensor_data = False

            shared_write_specs[-1].tensor_data_task = _FutureGroup(
                clone_tasks + clone_dependencies.futures
            )
            for w in shared_write_specs[:-1]:
                w.tensor_data_task = _FutureGroup(clone_tasks)
```

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+            )
+
+            clone_tasks: List[_Future] = []
+            for w in shared_write_specs[1:]:
+                clone_tasks.append(
+                    self._computation_pool.submit(
+                        self._do_clone, w, clone_dependencies
+                    )
+                )
+                w.user_owns_tensor_data = False
```
Eta0, line 4279: Adding a clone task can be skipped if a tensor is being cloned for not having been contiguous. Since that happens asynchronously, it should still be included up to this point (so its cloning status is tracked by `clone_dependencies`); it just won't need the final cloning part here.

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+
+        # If any tensors are shared, so we need to clone all but one of them before encrypting
+        write_specs_by_addr: Dict[int, List[TensorSerializer._WriteSpec]] = (
+            collections.defaultdict(list)
+        )
+        for w in write_specs:
+            if w.tensor.device.type != "cpu":
+                continue
+            address = w.tensor.untyped_storage().data_ptr()
+            write_specs_by_addr[address].append(w)
```
Eta0, line 4258: Interestingly, PyTorch counts tensors imported from other libraries like `numpy` as always having separate storages, even if they overlap or represent a completely identical region when compared with other storages—even in its `Tensor.is_set_to` function, so non-PyTorch-native tensors are sort of a "known bug" for this method, though it's niche enough to not affect almost any normal workflows.

Other than that, we can identify overlapping input data more precisely using an $O \lparen n \cdot \log \lparen n \rparen \rparen$ (where $n$ is the number of tensors) algorithm I designed for it once, since this old way incorrectly identifies disjoint slices of a tensor as being shared, but we can add that in another PR, since it's not a regression of this one.

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+    def _do_clone(write_spec, dependency: Optional[_Future]):
+        if dependency is not None:
+            dependency.result(_TIMEOUT)
+        write_spec.tensor = write_spec.tensor.clone().detach()
+        write_spec.numpy_tensor = _NumpyTensor.from_tensor(write_spec.tensor)
+
+    def _prepare_for_write_encryption(
+        self, write_specs: Sequence[_WriteSpec]
+    ) -> None:
+        assert self._encrypted and self._encryption is not None
```
Eta0, line 4248: I'd try to avoid `assert` statements for anything regarding encryption correctness (or anything else important) since `assert` statements can be skipped when optimization is enabled, just in case.
```suggestion
        if not self._encrypted or self._encryption is None:
            raise RuntimeError(
                "Tried to encrypt tensors without encryption parameters"
                " having been provided"
            )
```

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+
+            if w.data_length == 0:
+                # All headers are expected to have crypt_info segments, so add
+                # an empty one
+                w.crypt_info = _crypt_info.CryptInfo()
+                continue
+
+            if w.tensor_data_task is not None:
+                w.tensor_data_task.result(_TIMEOUT)
+                w.tensor_data_task = None
```
Eta0, line 4298: This should not need to block the main thread on `tensor_data_task`. Do you need me to change the `_crypt.ChunkedEncryption` interface to make this possible? I can probably make an option to provide the buffer after constructing it, as long as the length is provided in advance.

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+                w.tensor_type,
+                w.name.serialized_(),  # name as bytes
+                dtype_bytes,
+                w.shape,
+                w.data_length,
+                0,  # bogus file_offset. This gets filled in in build()
+                include_crc32=w.include_crc32,
+                include_sha256=w.include_sha256,
+                crypt_info=w.crypt_info,
+            )
```
Eta0, line 4347: The comment `# bogus file_offset. This gets filled in in build()` isn't accurate; it just gets filled in a bit later in this function but before `build()`. It's the tensor data offset that gets filled in in `build()`.
```suggestion
            w.header = _TensorHeaderSerializer(
                w.module_index,
                w.tensor_type,
                w.name.serialized_(),  # name as bytes
                dtype_bytes,
                w.shape,
                w.data_length,
                0,  # placeholder file_offset
                include_crc32=w.include_crc32,
                include_sha256=w.include_sha256,
                crypt_info=w.crypt_info,
            )
```

tensorizer/serialization.py
Eta0, line 1: This is more of a general comment as it doesn't block this PR, but as an optimization we should operate on detached copies of tensors to better control tensor lifetime and their garbage collection impact (by manually `set_`-ing them to empty buffers at the end of the serialization process). The handling of futures in this PR makes the object lifetime a little unclear, but depending on what function deallocation triggers in, it can reduce performance by a bit. Plus, deallocation can even be shunted to a background thread right before returning from the function that way.

Eta0, line 1: This one isn't a blocker for merging this PR, but an extra thought (doesn't require an answer either): Does queueing operations in batches to repeatedly loop over the entire list of tensors rather than perform many different operations in order on a each tensor negatively impact performance due to the worse memory access pattern? For small data it would, but I don't know if tensors are so large that memory caching wouldn't affect them much anyway. Hashing performance, for example, tends to improve when a tensor is split into chunks and alternating hashes are performed on the chunks, so I'm wondering how significant it might be to be able to fit operations like "hash a block of data, encrypt it, and write it" as back-to-back steps at some point. Defining compound operations like `stream_decrypt` was helpful in the deserializer, so something like that might be able to work in this organizational scheme too.

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+        for w in write_specs:
+            tensor_dtype_str = str(w.tensor.dtype)
+            if _NumpyTensor._is_asymmetric(w.tensor.dtype):
+                # is opaque
+                w.dtype = (
+                    f"<V{w.tensor.element_size():d}"
+                    + OPAQUE_DTYPE_SEP
+                    + tensor_dtype_str
+                )
+                w.set_min_file_version_number(OPAQUE_TENSORIZER_VERSION)
```
Eta0, line 4224: Does this make sense to include if all new serialized files are past this version anyway? Or are we planning on supporting serializing with older formats and potentially sharing this code path?

tensorizer/serialization.py
```diff
@@ -1674,6 +1710,31 @@ def __init__(
...
+                    self._file.seek(entry.offset)
+                    header = _TensorHeaderDeserializer.from_io(
+                        version_number,
+                        self._file,
+                        zero_hashes=True,
+                        check_crypt_info=self._has_crypt_info,
+                    )
+                    if header is None:
+                        raise KeyError("Unexpected empty header")
+                    self._headers[entry.name] = header
```
Eta0, line 1734: We should probably read the entire metadata section including these headers in one or two `read` (or `readv`) operations and just use the `_TensorHeaderDeserializer` buffer constructor instead of `from_io`.

tests/test_serialization.py
```diff
@@ -811,6 +811,113 @@ def try_serialize(obj):
                     self.assertNotIn("11", deserialized)
 
 
+class TestIncrementalSerialization(unittest.TestCase):
+    # Tests that involve multiple calls to write_tensor()
```
Eta0, line 815: It used to be the case that a bulk write could still be incremental, and a series of bulk writes could be used to create a tensorizer file. Bulk write functions are also the only way to incrementally write structured data, namely through `write_state_dict`. As far as I can tell, the way the serialization code is written still allows that, so we should add some test cases for that here too.

Eta0 (changes requested):
# Design Considerations
Separate from the issues in the comments, there are three points that require the most significant changes or design consideration to accompany this. Most of these are not "these need to be implemented in this PR" but "the designs we will need for these should be considered to see if the changes we are making now will be compatible with them," and is more for discussion.

## Accounting for Header Encryption
Tensorizer v3 should extend encryption to include header information. This needs to be accounted for in the file metadata and the patterns we use to buffer and write the metadata.
Important things to consider:
1. Where do we put encryption parameters in the header?
	- My take: Indicate in the file flags if the headers are encrypted, and then place encryption info immediately before the metadata section if so. It likely shouldn't be bundled into the metadata section, so that the length of the metadata section can also be encrypted. File flags and other meta-metadata like hashes of the metadata section can be encrypted too.
2. How does incremental writing interact with metadata encryption?
	- How many nonces and MACs are used when encrypting and decrypting the metadata section?
	- If one is used, then the metadata section can only be updated by rewriting the entire thing, because data can't be inserted into the middle of an encrypted stream (for the non-tensor-header metadata section), only appended at the end.
		- This might be cheap enough that we could do it anyway, though it feels like it could be limiting in the future.
	- If two are used (one for metadata entries, one for tensor headers), the metadata section could actually be written as two independent streams, only rewriting their MACs on each synchronization.
	- If multiple are used, then we need to handle size limits for the nonce & MAC list like size limits of the rest of the metadata section. (Luckily, this fits pretty well, since the size taken up by nonces and MACs is directly proportional to the number of entries being described when using one encrypted segment per metadata entry).
	- Ideally, information like the choice of encryption method would be stored in a tagged and extensible format like tensor `CryptInfo` segments, which would allow changing this in order to make it more or less rigid in the future.
3. To what extent do we attempt to protect information about model "size"?
	- The filesize gives away most information about that regardless of what we encrypt, but should we set out to protect information about how many tensors are contained in a file?
	- The number of tensors could potentially be leaked by information like the length of the metadata section, or the spacing of padding between tensor data entries. The length of the metadata section can be determined by the padding at the end of the metadata section unless we scramble the padding bytes during encryption.

We don't need to have functional header encryption to merge this PR, but deciding how it will fit into the metadata will help this one not conflict with that change when we do it later. I can work on the implementation for it later once it's more figured out.

## Accounting for File Range Insertion
A few serialization media allow inserting more space into the middle of a file instead of just at the end. In particular, `ext4` allows it, which accounts for much of local file serialization, and serialization to temporary files on a local disk, which accounts for object storage and (potentially) Azure blobstore serialization. If we eventually support object storage uploads without temporary files, those would similarly not limit adding more data into the middle of a file. For these cases we can do any header modifications we want without worrying about determining the required metadata size before writing tensor data, or we could even shrink over-allocated space in the metadata section to match what was written (deleting a file range).

For concerns about supporting this:
- Most filesystems only support adding new data ranges at logical block boundaries, so the metadata entry and header entry sections need to be separated by a page boundary when doing this.
	- They don't need to be separated that far in general, so this can be decided on-the-fly as part of the implementation, which ought to be compatible with this design, to my understanding.
- We could write data offsets into headers relative to the beginning of the data section of the file, so that we don't need to know the length of the metadata section in order to write headers. The beginning of the data section of the file ought to be known to the deserializer at the point of reading those anyway, so it could add that to each offset as soon as each header is deserialized, and there wouldn't need to be a significant amount of code changed in the deserializer.

## More Granular Future Dependencies
In this PR's changes, dependencies for later tasks are tracked as a single `tensor_data_task`. We should add differentiation for write hazards and read hazards, since storing it as one item doesn't allow two operations that only need to read tensor data to happen simultaneously (e.g. computing hashes while also writing data to disk, when encryption isn't active). This is less harmful to performance when queueing operations in batch order like now rather than per-tensor, but any time that there are enough threads (or few enough tensors) to handle tasks from multiple stages at once, this would unblock the later stages sooner.

In addition to those three things, there was also this optional suggestion, which would complement the first two somewhat:
## Optional Syncing
We could add a `synchronize=False` mode (referring to syncing data to disk) for incremental writes that defers writing headers until a write occurs with `synchronize=True` or the file is finalized. For cases like uploading to a temporary file before uploading to S3, writing headers can always be deferred, and this could be better for performance whenever constant synchronization isn't needed since it can guarantee a single write for that entire section, and (potentially) allow less fiddling with encryption information in the header section from having to repeatedly edit any data in it (like MACs).

## Future Work
Finally, these are various things on my v3 wishlist that aren't directly relevant to this PR and aren't yet implemented (nor should they be addressed here), but I wanted to make a note of them here for future reference.
- Add cryptographic authentication / checksums / ECC (if we're getting fancy) for the entire metadata and header section of the file, as a unit, since this allows checking that the metadata is complete and not missing entries (which could potentially happen if, say, the metadata length segment wasn't set properly at serialization time, or ended up changing somehow).
- Use more consistent and documented exception types
	- E.g. some sort of `InvalidTensorizerFile` exception when deserializing, which could have subtypes like `UnsupportedTensorizerVersion`
	- This would just make it a bit easier for code using tensorizer to differentiate between bad files or issues with their system, library dependency versions, deserialization flags they requested, or various other things
	- Having a smaller, documented set of errors in the outer layer of the API makes it so people don't have to use `except Exception` as often, which is usually nice
- Remove the `compress_tensors` option to the serializer and its vestigial related machinery, since it has done nothing even as far back as tensorizer v1.0.0
- Turn `TensorSerializer` into a context manager to match with `TensorDeserializer`
- I need to tweak `ChunkedEncryption` slightly in a way that is incompatible with older deserializers, so I'd like to add it in the new data version 5 alongside the changes in this PR

tensorizer/serialization.py
```diff
@@ -481,46 +509,46 @@ def __init__(
...
+            self.dtype,  # Dtype
+            self.shape_len,  # Shape length
+            *self.shape,  # Shape
+            self.file_offset,  # Header start (relative to the file)
             # Tensor data start (relative to the file):
-            file_offset + self.data_offset,
-            data_length,  # Tensor length
+            self.data_offset,
+            self.data_length,  # Tensor length
+        )
```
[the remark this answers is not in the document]
bchess: imo metadata is a dead-end. Keeping it intact for now makes it easier to load 2.x tensors, but in future major versions everything would be in header. Organizationally they are already the same as headers in 3.x

tensorizer/serialization.py
```diff
@@ -4446,17 +4128,492 @@ def write_state_dict(self, state_dict: Union[Dict, List, Tuple]):
...
+        self, write_specs: Sequence[_WriteSpec]
+    ) -> None:
+        def compute_crc32(
+            write_spec: TensorSerializer._WriteSpec,
+            dependency: Optional[_Future],
+        ):
+            if dependency is not None:
+                dependency.result(_TIMEOUT)
+            header_crc32 = write_spec.header.compute_crc32()
+            crc32 = zlib.crc32(write_spec.tensor_memoryview, header_crc32)
```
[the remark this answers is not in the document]
bchess: Agreed!

tensorizer/serialization.py
```diff
@@ -3700,529 +3766,152 @@ def write_tensor(
...
-            nonces = self._new_nonces(chunks.count)
-            encryptor = _crypt.ChunkedEncryption(
-                key=self._encryption.key,
-                buffer=tensor_memory,
-                chunk_size=self._crypt_chunk_size,
-                nonces=nonces,
-                executor=self._computation_pool,
+            self.metadata_pos = -1  # Set in _prepare_for_write_headers
+            self.encryptor: Optional[_crypt.ChunkedEncryption] = (
+                None  # $et in _do_encryption if encrypted
```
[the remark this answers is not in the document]
bchess: 💵 💵

tensorizer/serialization.py
bchess: I don't know, but my intuition is that it's minimal impact. Looping over tensors in aggregate makes it easier to handle batched operations or relationships/dependencies between tensors though.

tensorizer/serialization.py
bchess: Yeah its fine to remove

tensorizer/serialization.py
bchess: I don't really know why this was useful imo, but open to pushback. In a major version update we can break the interface

tensorizer/serialization.py
bchess: Yeah agreed there should be some docs on this

tensorizer/serialization.py
bchess: Updated!

tensorizer/serialization.py
bchess: Right now refer to my info in the PR description itself, but we should include something like that in the code

bchess:
> ## Accounting for File Range Insertion

Imo its a fair bit of added code complexity, adds more constraints to what we can optimize, for little benefit. Object stores upload APIs have minimums to the chunk sizes, and you still need to assign the correct index of the block ahead of time -- you can't insert a block later.

tensorizer/serialization.py
```diff
@@ -3700,529 +3766,152 @@ def write_tensor(
...
-                            new_cpu_tensor.detach(),
-                            timeout=_TIMEOUT,
-                        )
-                finally:
-                    # Sentinel
-                    transferred.put(None)
-                    transfer_finished = True
+        write_dependency: Optional[concurrent.futures.Future] = None
+        if not incremental:
+            write_dependency = self._maybe_fallocate(write_specs)
```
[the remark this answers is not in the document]
bchess: Yeah, but the logic inside `_maybe_fallocate()` uses its `tensors` list to calculate total size. To work in incremental mode it'd need to be updated to account for past writes.

tensorizer/serialization.py
```diff
@@ -3700,529 +3766,152 @@ def write_tensor(
...
-                interrupt_transfer,
-            ) = self._async_bulk_device_to_host_transfer(cuda_tensors)
-        else:
-            transferred = interrupt_transfer = None
-        del cuda_tensors
+            self._synchronize_pools()
+            self._sync_prologue_state()
+        except Exception as e:
+            for j in self._jobs:
+                j.cancel()
```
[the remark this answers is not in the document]
bchess: This is a good point. +1 to adding unit tests, because feels like something that could easily regress

tensorizer/serialization.py
bchess: The dependency itself is for `tensor_memory.nbytes` which then informs chunk count. I think you're right that the whole block could be pulled out into a thread, but it gets a little ugly that we'd start setting other attributes (`crypt_info` and `encryptor`) of `WriteSpec` on a thread without a more sophisticated mutex design. Correct me if I'm wrong, but I don't think there's anything here that's CPU intensive and going to be able to release the GIL. And we'd need all to reconcile by the next step of `_prepare_for_write_headers` anyway.

tensorizer/serialization.py
bchess: Agreed!

bchess:
In terms of dev process:

There are a lot of good suggestions on changes and improvements, but this PR is already uncomfortably big. I think it's easier to make those changes as subsequent and individual PRs against the 3.x branch. I can take point on writing up issues on all of the pending comments so that they don't get lost. For some of the spicier ones, we can also use the issues as separate threads of discussion on each topic. How does that sound?

tensorizer/_futuregroup.py
```diff
@@ -0,0 +1,56 @@
...
+        return all(f.cancelled() for f in self.futures)
+
+    def running(self) -> bool:
+        return any(f.running() for f in self.futures)
+
+    def done(self) -> bool:
+        return all(f.done() for f in self.futures)
+
+    def result(self, timeout=None) -> None:
+        _future_wait_and_raise(self.futures, timeout=timeout)
```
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/141

tensorizer/serialization.py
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/143

tensorizer/serialization.py
```diff
@@ -481,46 +509,46 @@ def __init__(
...
+            self.tensor_type.value,  # Whether this is a parameter or a buffer
+            self.dtype_len,  # Dtype length
+            self.dtype,  # Dtype
+            self.shape_len,  # Shape length
+            *self.shape,  # Shape
+            self.file_offset,  # Header start (relative to the file)
             # Tensor data start (relative to the file):
-            file_offset + self.data_offset,
-            data_length,  # Tensor length
+            self.data_offset,
```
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/142

tensorizer/serialization.py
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/144

tensorizer/serialization.py
bchess: https://github.com/coreweave/tensorizer/issues/145

tensorizer/serialization.py
bchess: https://github.com/coreweave/tensorizer/issues/146

tensorizer/serialization.py
bchess: https://github.com/coreweave/tensorizer/issues/148

tensorizer/serialization.py
Eta0: Once upon a time, this could potentially progressively deallocate memory by consuming a module, but it was sort of at the whims of the garbage collector, and wouldn't really reduce peak RAM usage anyway, so it wasn't very useful. This used to be the default behaviour, and the first time I started engaging with the tensorizer repo was to help someone debug this behaviour, because it was unintuitive and messing their stuff up. I don't see too much value to keeping it, especially since parallelized bulk writing tends to keep references to tensors around for much longer.

If anyone does need progressive deallocation of memory, this would probably be better implemented as a generalized callback hook for when each tensor has finished being processed, which would let other code deallocate tensor memory *and actually know that it has happened so that they can start taking advantage of it*... among other use cases. Since this is a blocking interface, it's currently unlikely that anyone could use it more effectively than they could write `del m` after the call returns, so the workaround for any code currently using it just to free memory at some nonspecific point shouldn't be terribly inconvenient.

tests/test_serialization.py
bchess: https://github.com/coreweave/tensorizer/issues/149

tensorizer/serialization.py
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/150

tensorizer/serialization.py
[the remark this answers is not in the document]
Eta0: That's fair. The fixed format for metadata has been problematic before because e.g. `data_length` would be far more useful if split into `serialized_length` and `deserialized_length` to separately communicate the file layout and the amount of RAM/VRAM usage the deserializer will need to anticipate. If it is only used as a shortcut to navigate the more detailed header section that follows it, then I'd be equally fine with it being a mostly-dead section (or just removing it now from 3.0 files and keeping only its deserializer).

tensorizer/serialization.py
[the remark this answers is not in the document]
Eta0: `fallocate` extends the file relative to the current file position, so it doesn't need anything special to account for past writes if the file position is reliable; it can just progressively extend the file.

bchess:
> In terms of dev process:
> 
> There are a lot of good suggestions on changes and improvements, but this PR is already uncomfortably big. I think it's easier to make those changes as subsequent and individual PRs against the 3.x branch. I'd also like to empower others to be able to make contributions as well 😜
> 
> I can take point on writing up issues on all of the pending comments so that they don't get lost. For some of the spicier ones, we can also use the issues as separate threads of discussion on each topic. How does that sound?



> > ## Accounting for File Range Insertion
> 
> ~Imo its a fair bit of added code complexity, adds more constraints to what we can optimize, for little benefit. Object stores upload APIs have minimums to the chunk sizes, and you still need to assign the correct index of the block ahead of time -- you can't insert a block later.~
> 
> Thinking about this more, I maybe have presumed different from what you were proposing. I don't think we should add a public interface that allows for the mid-range insertion of a tensor. But I could see the benefit of using fallocate holes to optimize the writes of incremental tensors.

Ticketing this as https://github.com/coreweave/tensorizer/issues/159

bchess:
> > ## Accounting for File Range Insertion
> 
> ~Imo its a fair bit of added code complexity, adds more constraints to what we can optimize, for little benefit. Object stores upload APIs have minimums to the chunk sizes, and you still need to assign the correct index of the block ahead of time -- you can't insert a block later.~
> 
> Thinking about this more, I maybe have presumed different from what you were proposing. I don't think we should add a public interface that allows for the mid-range insertion of a tensor. But I could see the benefit of using fallocate holes to optimize the writes of incremental tensors.

Presuming this was what you intended, ticketed as https://github.com/coreweave/tensorizer/issues/159

tensorizer/serialization.py
Eta0: `nbytes` can be calculated up-front in the `_WriteSpec` constructor as `self.serialized_length = (tensor.element_size() * tensor.nelement()) & ~-(tensor.device.type == "meta")` (at least, in the absence of compression, which would complicate more things than this), so this shouldn't need to wait for anything to get that datapoint. Adding synchronization barriers introduces issues with parallel usage of system resources (though you recently enough read my whole spiel on that!), so it's best to avoid them if possible.

tensorizer/serialization.py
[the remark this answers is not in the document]
bchess: https://github.com/coreweave/tensorizer/issues/159

tensorizer/serialization.py
bchess: Good point!

bchess: https://github.com/coreweave/tensorizer/issues/160

Eta0:
> > > ## Accounting for File Range Insertion
> > 
> > 
> > ~Imo its a fair bit of added code complexity, adds more constraints to what we can optimize, for little benefit. Object stores upload APIs have minimums to the chunk sizes, and you still need to assign the correct index of the block ahead of time -- you can't insert a block later.~
> > Thinking about this more, I maybe have presumed different from what you were proposing. I don't think we should add a public interface that allows for the mid-range insertion of a tensor. But I could see the benefit of using fallocate holes to optimize the writes of incremental tensors.
> 
> Presuming this was what you intended, ticketed as #159

Yes—no, we don't need an interface that controls the actual tensor insertion point in the file—it would rather be useful for expanding the metadata section at the top of the file instead of going through hoops to determine its size or requiring people to specify `max_tensors` in advance. So I'm referring to setting up the entire metadata section as a chunk of indeterminate size, for object storage purposes.

Though, as a sidenote, you can add new chunks into the middle of a multi-part object storage upload, because part indices don't need to be consecutive (for anything compatible with [AWS's object storage](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html)):

> ### Parts upload
> When uploading a part, in addition to the upload ID, you must specify a part number. You can choose any part number between 1 and 10,000. A part number uniquely identifies a part and its position in the object you are uploading. The part number that you choose doesn’t need to be in a consecutive sequence (for example, it can be 1, 5, and 14). If you upload a new part using the same part number as a previously uploaded part, the previously uploaded part is overwritten.

If we wanted, we could make the file header (start at) chunk 1 and the data section start at chunk 100, and have room to optionally insert up to 99 extension chunks before committing an upload. But we wouldn't need to.

PR #132, opened 2024-04-30
[docs] Add documentation about how to do serialization in a separate process

bchess:
Write up a discussion of how to do serialization in a subprocess, with a few different examples.

### DIFF
--- a/README-subprocess-serialization.md
+++ b/README-subprocess-serialization.md
@@ -1,0 +1,192 @@
+# Tensorizer serialization via subprocess
+
+If you're using Tensorizer serialization to write checkpoints during training,
+you may want to run the serialization concurrently from your training code so
+that you can execute your next training step as quickly as possible.  And
+because of the Python GIL, it's better to do this in a separate process so that the
+serialization doesn't utilize any of the GIL that you'd otherwise use in your training code.
+
+Keep in mind that this is a way to achieve _concurrency_, not instant
+snapshotting. The tensors you are checkpointing still need to be kept in memory,
+unmodified, for the duration of the serialization process. (Though you may
+choose to copy them out of CUDA memory into CPU memory. These tradeoffs are
+discussed below.)
+
+Also refer to [PyTorch Multiprocessing best
+practices](https://pytorch.org/docs/stable/notes/multiprocessing.html) for more
+details about using PyTorch across processes
+
+
+## Warning about fork() and threads
+Be aware that Python `os.fork()` is often not a viable option, as it can be known to cause deadlocks if you have multiple threads. Python 3.12 and above
+will [issue a deprecation warning](https://github.com/python/cpython/pull/100229) if you attempt this. Some 3rd-party packages that rely on sockets or file descriptors may also not behave correctly when a process unexpectedly forks. 
+
+A subprocess (fork + exec) is generally safer, but you do not inherently get
+shared memory with the calling process. `multiprocessing` has two ways to create
+a child process: `spawn` or `forkserver`. `spawn` should always be safe.
+`forkserver` can be faster but safety depends on the behavior of modules at
+import time. See
+https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+for more details.
+
+## If starting from CUDA
+Presuming your tensors are in CUDA memory, there are a couple different options.
+
+### Option 1: Communicate the CUDA tensors directly
+CUDA tensors can be "shared" to a subprocess very efficiently since it's only communicating a pointer to device memory.
+
+Basically send the CUDA tensors over a `multiprocessing.Queue` to a subprocess that does the serialization. Ensure that the CUDA tensors remain in device memory until the serialization process finishes.
+
+```python
+import torch
+from tensorizer import TensorSerializer
+from transformers import AutoModelForCausalLM
+import torch.multiprocessing as mp
+
+def do_serialize(uri: str, model: torch.nn.Module):
+    serializer = TensorSerializer(uri)
+    serializer.write_module(model)
+    serializer.close()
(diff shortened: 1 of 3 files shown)


README-subprocess-serialization.md
```diff
@@ -0,0 +1,192 @@
...
+    main()
+```
+
+## If starting from CPU memory
+
+Tensors in CPU memory need to moved to shared memory to be communicated with a subprocess. PyTorch `multiprocessing` will do this transparently, but be aware
+that a memcpy occurs. You'll also need additional "surge" CPU memory during the duration of the copy of each tensor. PyTorch copies tensors serially, so you need additional memory equal to the size of your largest tensor. This is only used during the memcpy itself. The original non-shared memory is immediately freed (unless it is also in use by some other tensor)
+
+Depending on how you are constructing your CPU tensor, you may be able to preemptively `tensor.share_memory()` ahead of time, thus saving a memcpy when
+passing to the subprocess.
```
Eta0, line 150: We can explicitly note here:

> [!WARNING]
>
> The main process should avoid modifying tensors while they are being serialized from shared memory, to avoid corrupting the written file. If serializing *with encryption* from shared memory, tensors should additionally not be read again until serialization has finished, as encryption temporarily modifies tensors in-place.
>
> If concurrent modification or access is necessary, move the tensors out of shared memory and into a copy in the subprocess before serialization. This can be done in the same style shown for snapshotting CUDA tensors in a previous example.

README-subprocess-serialization.md
```diff
@@ -0,0 +1,192 @@
...
+https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+for more details.
+
+## If starting from CUDA
+Presuming your tensors are in CUDA memory, there are a couple different options.
+
+### Option 1: Communicate the CUDA tensors directly
+CUDA tensors can be "shared" to a subprocess very efficiently since it's only communicating a pointer to device memory.
+
+Basically send the CUDA tensors over a `multiprocessing.Queue` to a subprocess that does the serialization. Ensure that the CUDA tensors remain in device memory until the serialization process finishes.
```
Eta0, line 38: Extra note that tensors must remain unmodified, and, if using encryption, not read again until serialization finishes, the same as the warning for CPU shared memory.

README-subprocess-serialization.md
```diff
@@ -0,0 +1,192 @@
...
+if __name__ == '__main__':
+    main()
+```
+
+### Option 2: Snapshot CUDA tensors to CPU memory in subprocess before serializing
+
+Once the tensors are in CPU memory, they no longer need to occupy CUDA memory. But the tensors
+will now need to occupy CPU memory until they are fully serialized.
+
+Do this by calling `model.to('cpu')` immediately after sending to serializer.
```
Eta0, line 85:
```suggestion
Do this by calling `model.to("cpu")` immediately after sending to the serializer.
```

README-subprocess-serialization.md
Eta0, line 1: Should move this to `docs/subprocess-serialization.md`

Eta0 (changes requested):
Good content, can add a couple warnings, move the file, and maybe format the code snippets with `black`/`isort`

docs/subprocess-serialization.md
```diff
@@ -0,0 +1,198 @@
...
+if __name__ == '__main__':
+    main()
+```
+
+### Option 2: Snapshot CUDA tensors to CPU memory in subprocess before serializing
+
+Once the tensors are in CPU memory, they no longer need to occupy CUDA memory. But the tensors
+will now need to occupy CPU memory until they are fully serialized.
+
+Do this by calling `model.to("cpu")` immediately after sending to serializer.
```
Eta0, line 85: Missing an article
```suggestion
Do this by calling `model.to("cpu")` immediately after sending to the serializer.
```

docs/subprocess-serialization.md
```diff
@@ -0,0 +1,198 @@
...
+https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+for more details.
+
+## If starting from CUDA
+Presuming your tensors are in CUDA memory, there are a couple different options.
+
+### Option 1: Communicate the CUDA tensors directly
+CUDA tensors can be "shared" to a subprocess very efficiently since it's only communicating a pointer to device memory.
+
+Basically send the CUDA tensors over a `multiprocessing.Queue` to a subprocess that does the serialization. Ensure that the CUDA tensors remain **unmodified** in device memory until the serialization process finishes. Note if using encryption, the tensors may be encrypted-in-place so they must also not read again until serialization finishes.
```
Eta0, line 38: Wait I forgot, the prohibition on reading when using encryption only applies for CPU tensors, since when serializing from CUDA it only modifies the temporary CPU copy. We can drop that sentence. My bad.
```suggestion
Basically send the CUDA tensors over a `multiprocessing.Queue` to a subprocess that does the serialization. Ensure that the CUDA tensors remain **unmodified** in device memory until the serialization process finishes.
```

Eta0 (changes requested):
Two small changes including a mistake from my last review, whoops.

README-subprocess-serialization.md
Eta0: Yeah, just left a second review when I noticed that, my bad

docs/subprocess-serialization.md
bchess: lol jinx

README-subprocess-serialization.md
Eta0: The part about it needing to remain unmodified still applies though

Eta0 (approved):

PR #163, opened 2024-06-28
Ensure that data is decrypted in the event of an exception

### LINKED ISSUE #144: [3.0] Canceled serialization should still do in-place decryption
> @[Eta0](https://github.com/Eta0) Eta0 [last week](https://github.com/coreweave/tensorizer/pull/127#discussion_r1649538227)
In-place decryption during serialization is extremely important in the error path: it must be absolutely assured to not leave tensor data left encrypted in-place. Because of that, you can't cancel all futures on an exception: the decryption futures must proceed if any encryption has occurred, and they must be sure to trigger even if their other dependent futures have been cancelled.

> On a similar note, it needs to be guaranteed that self._maybe_decrypt_data can be called to queue up the decryption task in the first place, even if the main thread throws an exception at some point after calling self._prepare_for_write_encryption and queueing encryption but before the self._maybe_decrypt_data call would have happened normally. It would probably be good to add a field in _WriteSpec stating whether the tensor data is currently encrypted or not to handle cases where only some tensors were successfully encrypted before an error occurred, as decrypting them without them having been encrypted would also scramble their data.
> 
> We could probably add unit tests for this, too, e.g. by mocking the function that writes tensor data such that it always throws an exception, and then ensuring that the tensor data after serialization exits is the same as it was before it started.
> 

> 
> @[bchess](https://github.com/bchess) bchess [2 days ago](https://github.com/coreweave/tensorizer/pull/127#discussion_r1653635491)
> This is a good point. +1 to adding unit tests, because feels like something that could easily regress

bchess:
Closes #144 

I think this was in better shape than initially expected, because `_maybe_decrypt_data()` did its work in a `finally:` block. Anyway, added two test cases. One that covers the event of a pwrite() raising an exception, and the other of an exception occurring in the main write block before `_maybe_decrypt_data()` is executed. The latter test was failing, and changes to the exception handler have been made to pass.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -3832,6 +3832,9 @@ class TensorSerializer:
             # They are often chained from one step of the process to the next
             self.tensor_data_task: Optional[_Future] = None
 
+            # only used for tracking decryption during error handling
+            self.decrypt_task: Optional[_Future] = None
+
         @property
         def tensor_memoryview(self) -> memoryview:
             if not self.tensor.is_contiguous():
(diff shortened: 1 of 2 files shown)


Eta0 (changes requested):
The way that this handles subsequent errors during decryption causes a bit of exception spam since it will log an error from nearly every decryption task with a dependency that failed, and when already in the error path, it is likely that many dependencies failed. It also reports concerning-looking errors about failing to decrypt data if the data was never encrypted to begin with, but they are only logged (without tracebacks) and not handle-able. I opened #164 to address these two issues; it reduces the number of misleading errors, elevates real decryption errors to be top-level exceptions returned from `_bulk_write`, and adds more test cases. With those changes merged onto this branch I'd consider this good to merge.

Eta0 (approved):

Issue #159, opened 2024-06-27
[3.0] Use fallocate holes to optimize header size


bchess:
From a few different places in the comments of https://github.com/coreweave/tensorizer/pull/127

In incremental mode, in filesystems that support it, we should be able to use `fallocate()` as a means to insert additional metadata+headers, rather than having to pre-allocate as specified by `max_tensors`

Issue #150, opened 2024-06-27
[3.0] Use fallocate in incremental mode


bchess:
Regarding

```
        if not incremental:
            write_dependency = self._maybe_fallocate(write_specs)
```

> [Eta0](https://github.com/Eta0) [last week](https://github.com/coreweave/tensorizer/pull/127#discussion_r1649535863) • 
> As far as I can tell, this is the only time this flag is used, and it doesn't actually correspond to whether it is an incremental write or not, only whether it is from a call to write_tensor or not, so this flag could be renamed for clarity. That aside, for a large enough single tensor, it's still fine to call fallocate; it can proceed in the background during tensor and metadata preparation, including hash calculation.
> 
> @[bchess](https://github.com/bchess) bchess [2 days ago](https://github.com/coreweave/tensorizer/pull/127#discussion_r1653634822)
> Yeah, but the logic inside _maybe_fallocate() uses its tensors list to calculate total size. To work in incremental mode it'd need to be updated to account for past writes.

Eta0:
Copying another of my comments from that thread (https://github.com/coreweave/tensorizer/pull/127#discussion_r1657800179):
> `fallocate` extends the file relative to the current file position, so it doesn't need anything special to account for past writes if the file position is reliable; it can just progressively extend the file.

Issue #148, opened 2024-06-27
[3.0] Read all metadata in one read


bchess:
> [Eta0](https://github.com/Eta0) [last week](https://github.com/coreweave/tensorizer/pull/127#discussion_r1649548674) • 
We should probably read the entire metadata section, including these headers, in one or two read (or readv) operations and just use the _TensorHeaderDeserializer buffer constructor instead of from_io.

PR #170, opened 2024-07-18
feat(serialization): Store relative offsets for tensor data and headers

### LINKED ISSUE #142: [3.0] Make data_offset relative to the data section of the file
> [Eta0](https://github.com/Eta0) [last week](https://github.com/coreweave/tensorizer/pull/127#discussion_r1649530675)
> This should be relative to the data section of the file to make headers less dependent on the length of the metadata section, which could still be indeterminate when this would otherwise be ready to commit.

Eta0:
# Segment-relative offsets in headers

Closes #142. Helpful for #148, #150, and #159.

## Overview

Instead of storing the absolute file position of header entries and tensor data, this change stores their offsets relative to the beginning of their respective segment (e.g. the header segment, or the data segment). This makes it easier to create and edit various parts of a file's metadata independently from other parts of the metadata or the tensor data. Information about absolute offsets are instead stored on a segment-by-segment basis in a "layout" section of the metadata, preceding the metadata and header sections.

This separates the file header and tensor data enough that it should even be possible now to split them into two completely separate files, if that design may be helpful later on.

This change also aligns the end of the metadata segment to a 4096-byte block boundary. The header segment also ends on a 4096-byte boundary as was the case before. This allows these segments to more easily be manipulated with `fallocate` for resolving #150 and #159 later.

## Deserialization

This change requires minimal modifications to deserialization. In newer files, the layout section is read, and the absolute offsets listed in it are used to convert all subsequently-read relative offsets into absolute offsets.

## Structure

The binary format of the layout section is:
- Byte length of the layout section (64-bit)
- Version tag (8-bit, always `1` for now)
- Number of segments (8-bit)
- Metadata segment
  - Tag (8-bit, always `1`)
  - Absolute start offset (64-bit)
  - Absolute end offset (64-bit)
- Header segment
  - Tag (8-bit, always `2`)
  - Absolute start offset (64-bit)
  - Absolute end offset (64-bit)
- Tensor data segment
  - Tag (8-bit, always `3`)
  - Absolute start offset (64-bit)

The tag fields aren't strictly necessary, but are present to make it slightly easier to write a deserializer that could handle the metadata segment being removed in the future, or optional segments, or variations of segments that include more or less information. I'm open to suggestions on improving the format for the layout section. Other potentially odd choices at present:

1. End offsets refer to the allocated space for the segment, including padding, and do not represent the length of meaningful content in a segment
  - Length of meaningful content may be encoded elsewhere in the metadata as appropriate
2. End offsets are included for the metadata and header segments, even though they are *currently* equal to the start offset of the following segment
  - This is here for convenience and in case segments somehow need to not be written back-to-back in a single file at some point
3. Tensor data doesn't have an end offset listed
 - The end of the tensor data segment is assumed to be the end of the file
 - This lessens the need to update the layout segment after every tensor (bulk-)write
 - We could probably add an end offset anyway if we wanted to since that information is available to the serializer during each prologue synchronization

If we want to support serialization to multiple files eventually, these fields could also be modified to include file identifiers, e.g. listing that the metadata is in file 0 but the tensor data segment is located in file 1, or potentially even something like listing multiple tensor data segments spread across multiple files.

## Other bug fixes

Buffered writes at the beginning of a file were not being properly flushed before switching to writing with `pwrite`, causing the beginning of a file to sometimes be overwritten with unfinished data. This was a regression from v2.9.0 where [the writer was always flushed after buffered writes](https://github.com/coreweave/tensorizer/blob/v2.9.0/tensorizer/serialization.py#L3445) and is fixed again in this PR.

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -321,6 +321,114 @@ class _FileHeader:
         return cls(version_number, feature_flags, tensor_size, tensor_count)
 
 
+@dataclasses.dataclass
+class _FileLayout:
+    _VERSION_TAG: ClassVar[int] = 1
+    _NUM_ENTRIES: ClassVar[int] = 3
+    metadata_start: Optional[int] = None
+    metadata_end: Optional[int] = None
+    _METADATA_TAG: ClassVar[int] = 1
+    header_start: Optional[int] = None
+    header_end: Optional[int] = None
+    _HEADER_TAG: ClassVar[int] = 2
+    data_start: Optional[int] = None
+    _DATA_TAG: ClassVar[int] = 3
+
+    _FORMAT: ClassVar[struct.Struct] = struct.Struct(
+        "<"
+        "B"  # Layout version tag
+        "B"  # Number of entries
+        "B"  # Metadata segment tag
+        "Q"  # Metadata segment start
+        "Q"  # Metadata segment end
+        "B"  # Header segment tag
+        "Q"  # Header segment start
+        "Q"  # Header segment end
+        "B"  # Tensor data segment tag
+        "Q"  # Tensor data segment start
+    )
+
+    _LENGTH_FORMAT: ClassVar[struct.Struct] = struct.Struct("<Q")
+    _LENGTH_SIZE: ClassVar[int] = _LENGTH_FORMAT.size
+    _FORMAT_SIZE: ClassVar[int] = _FORMAT.size
+    _FORMAT_SIZE_WITH_LENGTH: ClassVar[int] = _LENGTH_SIZE + _FORMAT_SIZE
+
+    @classmethod
+    def placeholder(cls) -> bytearray:
+        buffer: bytearray = bytearray(cls._FORMAT_SIZE_WITH_LENGTH)
+        cls._LENGTH_FORMAT.pack_into(buffer, 0, cls._FORMAT_SIZE)
+        return buffer
+
+    def pack(self) -> bytearray:
+        if None in (
+            self.metadata_start,
+            self.metadata_end,
+            self.header_start,
+            self.header_end,
+            self.data_start,
+        ):
+            raise RuntimeError("Cannot build; incomplete data")
+
+        buffer: bytearray = self.placeholder()
+
+        self._FORMAT.pack_into(
+            buffer,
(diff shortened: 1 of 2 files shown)


tensorizer/serialization.py
```diff
@@ -3502,26 +3614,87 @@ def __init__(
...
-        self._tensor_cur = (
-            self._header_end
-        )  # is the start of where to write tensor data. or None
+            approx_metadata_size: int = 256 * max_tensors
+            approx_header_size: int = 1024 * max_tensors
+
+            self._metadata_end = self._metadata_start + 8 + approx_metadata_size
+            # Extend the metadata segment to end on a block boundary
+            # This allows later manipulation with the fallocate syscall
+            # on coöperating operating systems to extend this segment
```
wbrown, line 3647: This spelling of `cooperating` intended? :)

tensorizer/serialization.py
```diff
@@ -3502,26 +3614,87 @@ def __init__(
...
+            # in the middle of the file
+            self._metadata_end -= self._metadata_end % -4096
```
wbrown, line 3649: Should block boundaries be a constant? i.e. do we know that the aligned boundaries are the same on ARM vs x86-64?

tensorizer/serialization.py
```diff
@@ -321,6 +321,114 @@ def from_io(
...
+    header_start: Optional[int] = None
+    header_end: Optional[int] = None
+    _HEADER_TAG: ClassVar[int] = 2
+    data_start: Optional[int] = None
+    _DATA_TAG: ClassVar[int] = 3
+
+    _FORMAT: ClassVar[struct.Struct] = struct.Struct(
+        "<"
+        "B"  # Layout version tag
+        "B"  # Number of entries
```
wbrown, line 340: This is the number of metadata header entries, which is currently `3`, correct? I ask because we have this as `B`. :)

wbrown (commented):
This looks good! :+1:

My general comment would be that we should probably encode a `data_length` field. As you mentioned in your PR description, `end_offset` no longer meaningfully tells us the actual length of the data.

We may want to make it easier on ourselves to slice that data by having the actual data length available.

PR #177, opened 2024-09-11
Support s3 client as an argument

angerrp:
**Problem:**
The current s3 parameters do not support all of the different possibilities of s3 authentication. As an example, the current authentication does not support aws session tokens which can prevent successful authentication. Further parameters which are not included can be found in the boto3 configuration e.g. the signature version. 
Introducing all of these parameters and configurations in `tensorizer` via `.s3cfg` or arguments to `tensorizer` would require a lot of effort and would be a maintenance burden.

**Proposal:**
This PR instead proposes to pass an already configured s3 client object to the `tensorizer` API, making the configuration of the s3 authentication more flexible.


angerrp:
Closing this as a botocore client is not pickleable and it would break a bunch of stuff

PR #181, opened 2024-11-28
fix(serialization): Slice 1-D multibyte data as bytes for `pwrite`

Eta0:
# 1-D Multibyte Data Slicing

This change fixes a bug in `TensorSerializer`'s write routine that would incorrectly slice 1-D multibyte data using its original `dtype` width rather than by individual bytes. The effect of this was that if attempting to write a buffer only wrote part of it, slicing the beginning of the buffer away to then attempt to write the rest of it would create an invalid slice, and then fail the write routine with an exception.

For example, if writing a 32-byte buffer of 4x `float64` elements partially succeeded and wrote 12 bytes, it would previously essentially attempt to then continue by writing `buffer[12:]`, but mistakenly skip ahead by the width of 12x `float64` elements (i.e. 96 bytes) instead of 12 bytes. It would then find that it had no more data to write, and then check that it wrote everything, and fail. To fix this, it now casts buffers to an unsigned byte type before slicing if it is not one already.

This only affects 1-D data because multidimensional arrays passed to `TensorSerializer`'s write routine were already correctly converted to 1-D arrays of bytes. Only ones that were already 1-D, but the wrong width, were missed.

## Code version update to v2.9.1

This change additionally updates the code version from v2.9.0 to v2.9.1 so that this fix can be published as a bugfix release.


tensorizer/serialization.py
```diff
@@ -3526,13 +3526,15 @@ def _pwrite_syscall(
             verify if isinstance(verify, int) else self._buffer_size(data)
         )
         bytes_just_written: int = os.pwrite(self._fd, data, offset)
-        bytes_written += bytes_just_written
+        if bytes_just_written > 0:
+            bytes_written += bytes_just_written
```
wbrown, line 3530: Hmm. Usually, that's `0` or greater. Under what situations does `pwrite` return a negative number?

tensorizer/serialization.py
```diff
@@ -3510,7 +3510,7 @@ def _mv_suffix(data: "collections.abc.Buffer", start: int):
         if not isinstance(data, memoryview):
             data = memoryview(data)
         try:
-            if data.ndim != 1:
+            if data.ndim != 1 or data.format != "B":
```
wbrown, line 3513: Would casting a byte format to a byte format be a no-op, effectively?

wbrown (commented):
I have questions!

tensorizer/serialization.py
Eta0: `pwrite` returns a negative number if it errored. This should hopefully be caught in CPython's wrapper around the actual syscall and converted to a Python exception, but just in case, it won't add it to the accumulator either.

tensorizer/serialization.py
Eta0: It would create an intermediate `memoryview` object that is referenced only by the one returned by this function. So it's not quite a no-op, but the output would be able to be used in the same way.

PR #182, opened 2025-01-17
support fp8 quantized models

btakeya:


btakeya:
Sorry for confusion. It's mistake that handling my internal work.

PR #183, opened 2025-01-25
Fix numpy deprecation product → prod

maresb:
`numpy.product` was deprecated in numpy 1.25, with this [PR](https://github.com/numpy/numpy/pull/23314) and removed in numpy 2.x.


maresb:
This identical fix was reimplemented yesterday in #184.

PR #185, opened 2025-03-07
Fix S3 Client Credential Handling and Initialization

Prikshit7766:
This PR improves S3 client initialization with credential handling and configuration. 
Reference: #177.


Eta0:
We've improved the handling for configuration detection from config files and environment variables in #189, which should cover this case as well.

PR #187, opened 2025-05-09
fix(stream_io): Default to signature version 4 for most object storage endpoints

Eta0:
# Default to signature version 4 for object storage reads

This change switches the default signature version for object storage reads from v2 to v4 when one is not specified explicitly for all endpoints other than `accel-object` ones. This is important because v2 signatures are not supported by all modern object storage backends.

This change also updates the CI workflows to use up-to-date runners, actions, and software versions.


wbrown (approved):
:+1:

PR #189, opened 2025-05-29
feat(stream_io): Improve object storage config handling

Eta0:
# Object storage config handling improvements & improved CAIOS support

These changes improve the compatibility of the `stream_io.open_stream()` object storage feature with different types of object storage configurations.

### DIFF
--- a/tensorizer/stream_io.py
+++ b/tensorizer/stream_io.py
@@ -897,6 +897,11 @@ def _ensure_https_endpoint(endpoint: str):
         raise ValueError("Non-HTTPS endpoint URLs are not allowed.")
 
 
+def _is_caios(endpoint: str) -> bool:
+    host = urlparse(endpoint if "//" in endpoint else "//" + endpoint).netloc
+    return host.lower() in {"cwobject.com", "cwlota.com"}
+
+
 def _new_s3_client(
     s3_access_key_id: str,
     s3_secret_access_key: str,
@@ -914,6 +919,9 @@ def _new_s3_client(
     config_args = dict(user_agent=_BOTO_USER_AGENT)
     auth_args = {}
 
+    if _is_caios(s3_endpoint):
+        # These endpoints don't support path-style addressing
+        config_args["s3"] = {"addressing_style": "virtual"}
     if s3_region_name is not None:
         config_args["region_name"] = s3_region_name
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+[Unreleased]
+
+### Fixed
+
+- `stream_io.open_stream()` now uses virtual-hosted-style
+  bucket addressing for the `cwobject.com` and `cwlota.com` endpoints
+
 [2.9.3] - 2025-05-09
 
 ### Changed
@@ -424,6 +431,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `get_gpu_name`
   - `no_init_or_tensor`
 
+[Unreleased]: https://github.com/coreweave/tensorizer/compare/v2.9.3...HEAD
 [2.9.3]: https://github.com/coreweave/tensorizer/compare/v2.9.2...v2.9.3
 [2.9.2]: https://github.com/coreweave/tensorizer/compare/v2.9.1...v2.9.2
 [2.9.1]: https://github.com/coreweave/tensorizer/compare/v2.9.0...v2.9.1


sangstar (approved):
LGTM. Test cases passing on my end.

tensorizer/stream_io.py
```diff
@@ -897,23 +908,24 @@ def _ensure_https_endpoint(endpoint: str):
         raise ValueError("Non-HTTPS endpoint URLs are not allowed.")
 
 
+def _is_caios(endpoint: str) -> bool:
+    host = urlparse(endpoint if "//" in endpoint else "//" + endpoint).netloc
+    return host.lower() in {"cwobject.com", "cwlota.com"}
```
arsenetar, line 913: Really not a fan of us hardcoding these like this (or the ord1 defaults we have had elsewhere for some time). But can be convinced this is ok.

For reference to those other hardcoded values that should be either A) dropped, or B) changed to our current object storage offering.
``` python
default_s3_read_endpoint = "accel-object.ord1.coreweave.com"
default_s3_write_endpoint = "object.ord1.coreweave.com"
```

tensorizer/stream_io.py
```diff
@@ -897,23 +908,24 @@ def _ensure_https_endpoint(endpoint: str):
...
-        raise TypeError("No secret key provided")
-    if s3_access_key_id is None:
-        raise TypeError("No access key provided")
-    if s3_endpoint is None:
-        raise TypeError("No S3 endpoint provided")
-
     config_args = dict(user_agent=_BOTO_USER_AGENT)
     auth_args = {}
 
+    if _is_caios(s3_endpoint):
```
arsenetar, line 926: This seems like something that should only be set if not already explicitly set by the configuration from boto3.

tensorizer/stream_io.py
```diff
@@ -1364,30 +1399,48 @@ def open_stream(
...
-                    f"\nReason: {e}"
-                    "\nIf the connection failed due to missing permissions"
-                    " (e.g. HTTP error 403), try providing credentials"
-                    " directly with the tensorizer.stream_io.open_stream()"
-                    " function."
-                )
-                s3_access_key_id = s3_access_key_id or ""
-                s3_secret_access_key = s3_access_key_id or ""
+
+            if force_http is None and s3.use_https is not None:
```
arsenetar, line 1403: This should follow after endpoint resolution as boto3 uses the `https://` or `http://` prefix on the configured endpoint respectively, which this logic should respect the same as the s3cmd's config version (which uses an explicit variable).

Essentially: https://github.com/boto/botocore/blob/a82eaae34b294970578e3af18d31242ac719ded0/botocore/session.py#L887C30-L892C36

tensorizer/stream_io.py
Eta0: I'm pretty sure in `boto3`'s source tree there is a large data segment that has these sorts of hardcoded compatibility checks for a huge set of different AWS endpoints, which is how Boto3's "magically pick the right one" feature works. Unless/until we upstream a change to support our object storage endpoints correctly, which I'm not sure they'd accept, this seems to me like a fair place to put a workaround.

The other hardcoded values are necessary for the default `s3://tensorized` bucket to "just work" since they're in that location. I'd like to change it to the newer endpoint in release 3.0.

tensorizer/stream_io.py
Eta0: It is something that can be set in an `.aws/config` file for Boto3, but since:

1. This is the only valid option for these endpoints anyway
2. (AFAIK) We don't provide users a generated `.aws/config` file or template to use in our web console nor in our CW documentation, or explain how to configure this in our CW documentation

It's good to ensure this here for these specific endpoints. The default behaviour in Boto3 is still used for other endpoints. Though we ought to update our documentation to clarify this as well.

tensorizer/stream_io.py
Eta0: Will fix this. Planning this logic:
- If `force_http` is specified → use HTTP, otherwise
- If `.s3cfg` has `use_https = False` → use HTTP, otherwise
- If the endpoint is resolved by `botocore` or `AWS_ENDPOINT_URL` environment variables → use the endpoint's scheme, otherwise
- Use HTTPS

tensorizer/stream_io.py
arsenetar: 1. That is fair, if boto3 has similar custom configurations then I have less pushback on this.
2. We actually do have our CW documentation https://docs.coreweave.com/docs/products/storage/object-storage/how-to/manage-buckets#manage-buckets-with-aws-cli.

tensorizer/stream_io.py
arsenetar: If this is a pattern already used in boto3 then probably fine... although we are not automatically changing the `force_http` based on if we are using `cwlota.com` or not which would fall into the same pattern as what we are doing here.  Same points apply on both sides there as well as if a user configures it correctly it will work, but by default with a minimal amount of configuring it will not.

Eta0: I am nervous about defaulting `force_http=True` for the `cwlota.com` endpoint on the basis that it could transmit a presigned URL as part of a request over an unknown network in cleartext if it is used somewhere where `cwlota.com` doesn't resolve to a local network (so not a CW datacenter, maybe from someone's own machine where they're trying to test something), although I think the presigned URL would only be valid for use by someone who can access `cwlota.com` endpoints. I'd rather leave that one specifically as something that needs to be declared explicitly in the config.

tensorizer/stream_io.py
arsenetar: Those docs leave a bit to be desired yes.

sangstar (dismissed):
Changes confirmed as working on my end once again, LGTM. Will defer to Andrew and his review comments.

tensorizer/stream_io.py
arsenetar: This seems fine.

Eta0: Implemented in 8dfb67d88a14c9e8a90cce131b0702141514e2e5.

arsenetar (approved):
Seems like this is behaving like I expect it to, granted not 100% validated e2e.

PR #192, opened 2025-06-27
fix(serialization): Detect and respect CUDA device contexts

Eta0:
# CUDA Device Contexts

This change fixes an issue where `TensorDeserializer` objects did not respect CUDA devices selected using context managers, e.g.:

```py
with torch.device("cuda:1"):
    with TensorDeserializer(file) as deserializer:
        ...  # Actually loads to cuda:0
```

More details can be found in the changelog, copied below.

## Changelog
- `TensorDeserializer` objects now respect CUDA devices chosen with the `torch.device()` and `torch.cuda.device()` context managers
  - Previously, the `device` parameter to the `TensorDeserializer` constructor was the only way to choose between multiple CUDA devices (e.g. `cuda:0`, `cuda:1`)
  - Now, when a device with no index is specified, such as `device=torch.device("cuda")`, `device="cuda"`, or even `device=None`, the local device context is used to select the device
  - `torch.cuda.set_device()` and `torch.set_default_device()` are also supported
  - This doesn't add support for detecting CPU contexts such as `torch.device("cpu")` without explicitly specifying `device="cpu"`
    - This is for backwards compatibility, as the `TensorDeserializer` default device has always been mandated to be CUDA whenever CUDA is available, and `torch` does not provide a public interface to reliably disambiguate an intentional `torch.device("cpu")` context from the global default
  - Lazy-loaded tensors will use the active device as of the time they are actually loaded
- `TensorDeserializer` objects no longer fail to open file-like objects whose `fileno()` methods raise `io.UnsupportedOperation`

## Lazy-loading
An interesting side effect of this change is that it is now possible to load to multiple devices using a single `TensorDeserializer`, as long as it is lazy-loaded. For example:

```py

cuda0_keys = { ... }
cuda1_keys = { ... }

with TensorDeserializer(file, lazy_load=True) as deserializer:
    with torch.device("cuda:0"):
        cuda0_tensors = deserializer.tree(filter_func=cuda0_keys.__contains__)
    with torch.device("cuda:1"):
        cuda1_tensors = deserializer.tree(filter_func=cuda1_keys.__contains__)

assert {t.device for t in cuda0_tensors.values()} == {torch.device("cuda:0")}
assert {t.device for t in cuda1_tensors.values()} == {torch.device("cuda:1")}
```

This is not *optimal* for loading to multiple devices, because it necessarily uses two bulk loads rather than one. To better support multi-device loading in the future, the `TensorDeserializer` loading logic should have a parameter to decide the destinations of each tensor—but that is unrelated to this PR.

## Precedence of device selections

PyTorch has multiple ways to set a default device, and the logic implemented in this PR does not match it 1:1 because it is pretty weird. This is documented in the `tensorizer.serialization._resolve_cuda_device` docstring, copied here:
> The precedence for CUDA device contexts here is:
>
> 1. `torch.device()` context managers
> 2. `torch.set_default_device()` defaults
> 3. `torch.cuda.current_device()` / `torch.cuda.device()` contexts
>
> This is slightly different from torch's own precedence, which is weirder:
>
> 1. If `device=None` in a function call:
>     1. `torch.device()` context managers
>     2. `torch.set_default_device()` defaults
> 2. If `device="cuda"` in a function call or from the previous step:
>     1. `torch.cuda.current_device()` / `torch.cuda.device()` contexts
>
> Note that in torch, `device="cuda"` ignores the local `torch.device()` context, but respects the local `torch.cuda.device()` context. This is not the case here, where `device="cuda"` respects either. Both in torch and here, if `device=None` but the local `torch.device()` or `torch.set_default_device()` contexts are `"cuda"` without an index, then the index is determined according to `torch.cuda.current_device()`, the same as if `device="cuda"` had been passed.

In summary, PyTorch's behaviour follows this pattern:

```py
with torch.device("cuda:1"):
    print(torch.empty(0).device)                 # "cuda:1"
    print(torch.empty(0, device="cuda").device)  # "cuda:0"

with torch.cuda.device("cuda:1"):
    print(torch.empty(0).device)                 # "cpu"
    print(torch.empty(0, device="cuda").device)  # "cuda:1"
```

Whereas in ours, combining contexts and `device="cuda"` or `device=None` parameters in any of these four ways are all equivalent (when CUDA is available) and will all load to `cuda:1`.


wbrown (dismissed):
Holy cow. The fact that lazy-loading can be used to redirect to multiple GPUs is pretty neat, but almost hacky. :)

Great job.

wbrown (approved):

PR #201, opened 2026-02-06
Add native multi-GPU device_map support to TensorDeserializer

abatilo:
Add native multi-GPU device_map support to TensorDeserializer

feat(serialization): add device_map parameter for multi-GPU tensor loading

Add native multi-GPU support to TensorDeserializer via a new device_map
keyword argument. Supports explicit per-tensor placement via a Mapping
and automatic greedy largest-first balancing via a Sequence of devices.

- Greedy balancer uses deserialized_length (bytes) with a min-heap
- Per-device CUDA streams cached in copy threads
- CPU tensors in mixed maps get dedicated buffers to avoid corruption
- Underspecified CUDA devices resolved up front
- read_numpy_arrays guarded against non-CPU device_map targets
- Default path (device_map=None) avoids per-tensor overhead in hot loop
- 19 new unit tests covering placement, balancing, fallback, edge cases

### DIFF
--- a/tensorizer/serialization.py
+++ b/tensorizer/serialization.py
@@ -40,6 +40,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    Mapping,
     Optional,
     Sequence,
     Set,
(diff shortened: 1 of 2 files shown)


abatilo:
This change is part of the following stack:

- #201 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

PR #203, opened 2026-04-16
Fix Constants() to work with Python 3.14 annotations.

nascheme:
In Python 3.14 (PEP 649), annotations are lazily evaluated and stored as `__annotate_func__` instead of `__annotations__`.  Add `_pop_annotations()` function to use in `tensorizer/_crypt/_encryption.py`.  Without this change, you get KeyError on `crypto_stream_salsa20_NONCEBYTES` when running with 3.14 (i.e. the annotation on the class cannot be found).

### DIFF
--- a/tensorizer/_crypt/_encryption.py
+++ b/tensorizer/_crypt/_encryption.py
@@ -346,6 +346,21 @@ crypto_stream_salsa20_xor_ic = init_crypto_stream_salsa20_xor_ic()
 sodium_memzero = init_sodium_memzero()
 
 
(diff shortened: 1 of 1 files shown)


Eta0:
Thank you for this contribution! I batched this in with a few other changes in #204 with you marked as a co-author. I did edit this slightly to use `annotationlib` to interface with the annotate function, though, as it seems accessing `__annotate_func__` directly is considered spicy (astral-sh/ruff#17859).

The fixed version will be available as `tensorizer` release 2.12.1.

